PR #1, opened 2025-02-08
Make User.sol/setTASigningKeyAddress an ownable function

mwittie:
This PR makes User.sol/setTASigningKeyAddress an ownable function. It also cleans up the instructions and dependencies for running the `on-chain` example.

### DIFF
--- a/on-chain/Makefile
+++ b/on-chain/Makefile
@@ -2,16 +2,11 @@
 test-local:
 	npx hardhat test --grep "Local"
 
-.PHONY: compile
-compile:
-	npx hardhat compile
-
-.PHONY: deploy
-deploy: compile
-	npx hardhat run scripts/deployUser.ts --network base-sepolia
-	npx hardhat verify --network base-sepolia $$(cat .user_deployed_address)
-
 .PHONY: copy-from-iterate
 copy-from-iterate:
 	mkdir -p inputs/
 	cp ../iterate/tmp/prev.json inputs/
+
+.PHONY: very-clean
+very-clean:
+	@rm -rf artifacts cache node_modules typechain-types
--- a/on-chain/README.md
+++ b/on-chain/README.md
@@ -1,46 +1,40 @@
 # On Chain
 
 This demo shows how we can get transitive attested data on chain. In this demo,
-bring a transitive attested function call from the [iterate demo](../iterate/)
+bring a transitive attested function call from the [iterate demo](../iterate)
 on to a chain in a development environment.
 
 ## Setup
 
-You should not do this again, but here are the steps documenting getting
-the project setup.
+Use `nix` to set up a shell environment with the correct dependencies:
 
 ```bash
-npm install npm@10.9.0
-npm init
-npm install --save-dev hardhat
-npx hardhat init
+nix-shell
 ```
 
-The last step may hang, you can cancel and reissue the command
+Continue with the following commands in the nix shell.
+
+Set up the project dependencies:
 
 ```bash
-npm install --save-dev "@nomicfoundation/hardhat-toolbox@^5.0.0" --dd
+npm install
 ```
 
-Install other dependencies
+## Run
 
-
-```bash
-npm install --save-dev -dd solidity-bytes-utils
-```
-
-You are ready to start going!!
-
-## Bring results from iterate on chain
-
-Copy an output from iterate
-
-```bash
-make copy-from-iterate
-```
-
-And run the test:
+Test bringing a transitive attested function call on chain into the
+[User](contracts/User.sol) contract:
 
 ```bash
 make test-local
+```
+
+You can see the details of this test in [test/User.ts](test/User.ts).
+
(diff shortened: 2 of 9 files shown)


on-chain/Makefile
```diff
@@ -2,16 +2,11 @@
...
-	npx hardhat verify --network base-sepolia $$(cat .user_deployed_address)
-
 .PHONY: copy-from-iterate
 copy-from-iterate:
 	mkdir -p inputs/
 	cp ../iterate/tmp/prev.json inputs/
+
+.PHONY: very-clean
+very-clean:
+	@rm -rf artifacts cache node_modules typechain-types
```
mwittie, line 12: Undoing the changes of `npm-install` and `make test-local`

on-chain/README.md
```diff
@@ -1,46 +1,40 @@
...
-You should not do this again, but here are the steps documenting getting
-the project setup.
+Use `nix` to set up a shell environment with the correct dependencies:
 
 ```bash
-npm install npm@10.9.0
-npm init
-npm install --save-dev hardhat
-npx hardhat init
+nix-shell
```
mwittie, line 12: I added this to make sure that the environment gets set up correctly

on-chain/contracts/TAParser.sol
```diff
@@ -1,4 +1,4 @@
-// SPDX-License-Identifier: UNLICENSED
+//  SPDX-License-Identifier: MIT
```
mwittie, line 1: Without this line hardhat throws warning. So I updated to MIT, since that's what we're using in the repo

on-chain/contracts/TAParser.sol
```diff
@@ -52,7 +52,7 @@ contract TAParser {
     {
         TA memory ta = decodeTA(taData);
 
-        FnCallClaims memory claims = decodeAPICallClaims(ta.Data);
+        FnCallClaims memory claims = decodeFnCallClaims(ta.Data);
```
mwittie, line 55: I udpated the function names since these are for `attest-fn-call` not `attest-api-call`

on-chain/contracts/User.sol
```diff
@@ -1,26 +1,30 @@
...
 pragma solidity ^0.8.10;
 
+import "@openzeppelin/contracts/access/Ownable.sol";
+import "hardhat/console.sol";
 import {JsmnSolLib} from "../lib/JsmnSolLib.sol";
 import {TAParser} from "./TAParser.sol";
 import {console} from "hardhat/console.sol";
 
-contract User is TAParser {
+contract User is Ownable, TAParser {
```
mwittie, line 10: Adding `Ownable` to `setTASigningKeyAddress`

on-chain/package.json
```diff
@@ -1,21 +1,20 @@
...
-    "dotenv": "^16.4.7",
     "npm": "^10.9.0"
   },
   "name": "on-chain",
   "version": "1.0.0",
-  "description": "You should not do this again, but here are the steps documenting the process.",
   "main": "index.js",
   "devDependencies": {
     "@nomicfoundation/hardhat-toolbox": "^5.0.0",
+    "@openzeppelin/contracts": "^5.2.0",
```
dlm, line 11: Are these dev dependencies?  I am still fuzzy on how npm + smart contracts differentiate dev vs prod dependencies, so what you have may be correct.  I'm really not sure.

on-chain/package.json
```diff
@@ -1,21 +1,20 @@
...
     "hardhat": "^2.22.18",
     "solidity-bytes-utils": "^0.8.3"
   },
   "scripts": {
     "test": "echo \"Error: no test specified\" && exit 1"
   },
-  "author": "",
-  "license": "ISC"
+  "author": "Blocky",
```
dlm, line 18: `Blocky, Inc.` ?

on-chain/shell.nix
```diff
@@ -8,7 +8,6 @@ in
 pkgs.mkShellNoCC {
   packages = with pkgs; [
     gnumake
-    nodejs_18
-    jq
```
dlm, line 12: thank you for removing!

on-chain/shell.nix
```diff
@@ -8,7 +8,6 @@ in
...
+    nodejs_20
```
dlm, line 11: Not a big deal, but any reason for the update?  It seemed okay on 18.  Did you add a v20 feature?

dlm (approved):
A few small comments/questions, but nothing significant.  :ship:

on-chain/package.json
mwittie: I ready about it seems that dev are the ones only used in test. I reorganized.

on-chain/shell.nix
mwittie: No, just what I had on my machine and I wanted to check if it removed some deprecation messages. Either way, but I figured using later is better. We can roll back if we need to.

PR #2, opened 2025-02-14
Add the fetch and process api example

mwittie:
This PR adds an example of using `attest-fn-call` to fetch and process data from the API.


fetch_and_process_api_call/fn-call.json
```diff
@@ -0,0 +1,13 @@
+[
+    {
+        "code_file": "tmp/x.wasm",
+        "function": "myOracleFunc",
+        "input": {
+            "market": "Binance",
+            "coin_id": "bitcoin"
+        },
+        "secret": {
+            "api_key": "CG-NqKT8uUEv9tfwG62w9EUqF7j"
```
dlm, line 10: Note that is my coin gecko key.  I was planning on rotating it, but, we should figure out how we want to handle this since I don't think we want to give this out to people.

fetch_and_process_api_call/as/verify.go
```diff
@@ -0,0 +1,53 @@
+package as
```
dlm, line 1: It looks like you are not using the verification tools in this demo and so this file should be able to be removed

fetch_and_process_api_call/as/result.go
```diff
@@ -0,0 +1 @@
+package as
```
dlm, line 1: you can probably remove this file

fetch_and_process_api_call/as/log.go
```diff
@@ -0,0 +1,29 @@
+package as
```
dlm, line 1: I only did a quick check, but it looks like you are not using logging in the example and so this could be deleted (or you could add some logging in to the example, which could be a nice touch)

fetch_and_process_api_call/Makefile
```diff
@@ -0,0 +1,16 @@
...
+tmp:
+	@mkdir -p tmp
+	
+tmp/x.wasm: main.go | tmp
+	tinygo build -o tmp/x.wasm -target=wasi main.go
+
+build: tmp/x.wasm
+
+run: build
+	cat fn-call.json | ./bky-as attest-fn-call  | jq -r .function_calls[0].claims.output | base64 -d | jq .
```
dlm, line 12: consider updating so that this doesn't require `bky-as` to be in the root of the directory.  In the quick examples I was creating, that worked, but for this one, it could be a pain.

consider updating so that the base64 decode is part of the first jq call so that you can avoid the call to `base64`.  Generally, on stuff I share, I tend to avoid using the base64 command unless it is needed because it is one of the commands that are different between gnu and bsd.

fetch_and_process_api_call/README.md
```diff
@@ -0,0 +1,31 @@
...
+
+This example shows how to use the Blocky Attestation Service (Blocky AS) to
+fetch data from an API and process it. The example uses the Coingecko API to
+fetch the current price of Bitcoin in USD on the Binance market.
+
+To get started, install the Blocky AS CLI by following the
+[setup instructions](https://blocky-docs.redocly.app/attestation-service/setup)
+in the Blocky AS documentation.
+
+Next, use `nix` to set up the correct environment for building WASM binaries
```
dlm, line 11: FWIW, one can avoid all the nix stuff by using tinygo via docker (that is what @matteobjornsson) does for his work environment.  That being said, personally, I think the nix version is less likely to run into errors, however, it is also more likely to have uses "fall off" while waiting for the environment to initialize on the first run.

fetch_and_process_api_call/shell.nix
```diff
@@ -0,0 +1,14 @@
+let
+  nixpkgs = fetchTarball "https://github.com/NixOS/nixpkgs/tarball/nixos-24.05";
```
dlm, line 2: This is fine for now, but maybe we should add a card to update the version of tinygo and the build flags.  V24.05 is getting close to a year old.  I wonder if this makes our examples look abandoned or dated.

fetch_and_process_api_call/README.md
mwittie: OK, I tried not using nix-shell and building with
```
docker run --rm -v .:/src -w /src tinygo/tinygo:0.35.0 tinygo build -o tmp/x.wasm -target=wasi main.go
```
the build works, but I'm getting runtime errors.

```
2025/02/13 19:06:38 Error: making fn call attestation request: making function call: unexpected status, wanted '200' got '422': "invoking code: error invoking code myOracleFunc: writing input to shared memory: allocating memory: allocating memory: module closed with exit_code(0)"
```

@matteo is there a better way to build? I could try tracking this down, but maybe you have this figured out already?

fetch_and_process_api_call/Makefile
mwittie: OK, I baked in base64 to the jq command string and added the current directory to the path before invoking `bky-as`

fetch_and_process_api_call/as/verify.go
mwittie: removing

fetch_and_process_api_call/README.md
matteobjornsson: Taking a look now

matteobjornsson: In the meantime, this is the script I've been using for my tinygo command:
```
#!/bin/bash

docker run --rm \
    -v "$(pwd)":/src \
    -w /src \
    --platform linux/arm64 \
    tinygo/tinygo:0.31.2 \
    tinygo "$@"
```

matteobjornsson: Two notes so far:
1. `nix-shell` took ten minutes. For some reason the tinygo package seemed to run a bunch of regression tests, I'm not sure what that's about. 
2. within nix-shell, `make build` did not work on my end due to however that c file is getting imported.
```
Building WASM module...
# github.com/blocky/as-demo/as
as/lib.go:8:13: fatal: 'stdlib.h' file not found
make: *** [Makefile:8: tmp/x.wasm] Error 1
```

matteobjornsson: Calling `make build` outside of the nix shell worked on my end, but that isn't as helpful for this example since we are trying to get end user set up

matteobjornsson: Also, regarding your docker command, last time I tried to build with tinygo 0.35 I got those errors, I think there are some breaking changes between 0.32 and 0.35 regarding build flags.

mwittie: Thanks

fetch_and_process_api_call/README.md
```diff
@@ -0,0 +1,39 @@
...
+  },
+  "Error": ""
+}
+```
+
+> Note: The `tmp/x.wasm` target in the `Makefiile` builds WASM for the
+> `linux/amd64` platform. That is the platform used by the hosted Blocky AS
+> server. If you're running Blocky AS locally, with `host = "local-server"` in
+> `config.yaml`, you may need to update the `tmp/x.wasm` target to build for 
+> your architecture.
```
mwittie, line 39: @matteobjornsson I'm guessing here based on the script you sent me. Is this true?

matteobjornsson: I don't think this is true. The script I use for tinygo just specifies details about the docker container and running the command via docker. All the details about the build target come from the flags in the Makefile. The `wasi` target is platform independent (in the context of linux/amd64 vs darwin/amd64, etc)

fetch_and_process_api_call/as/log.go
mwittie: reported a bug https://blocky.atlassian.net/browse/BKY-6321 and removed the file

fetch_and_process_api_call/shell.nix
mwittie: Removing, since we're using docker in this example

fetch_and_process_api_call/fn-call.json
mwittie: OK, I set it up as a place holder

fetch_and_process_api_call/README.md
mwittie: @matteobjornsson OK, I removed, could you check on your machine?

fetch_and_process_api_call/main.go
```diff
@@ -0,0 +1,129 @@
...
+	}
+
+	output := Output{Value: price, Error: errString, IsErr: isErr}
+	outputData, err := json.Marshal(output)
+	if err != nil {
+		// We panic on errors we cannot communicate back to function caller
+		panic("Fatal error: could not marshal output data")
+	}
+	return as.ShareWithHost(outputData)
+}
```
matteobjornsson, line 50: If you notice in `as/http.go`, the naming convention that's used there for "output" is the "Output" is the value we care about, and "result" is the struct that has the following format
```go
type Result struct {
	IsOk  bool   `json:"ok"`
	Value Output `json:"value"`
	Error string `json:"error"`
}
```
I don't think it really matters, but if you want to use such a naming convention, it would be consistent with the `as` package. In this file, you're using `Output` as a `Result` type. (note, these "naming conventions" are pretty fresh and in-flux, so one way vs the other is not necessarily "right", but might be where some of the ideas on the SDK are headed.)

Also, if you wanted to break up this `writeOutputToSharedMem` function, the cleanest implementation I have so far is in `test/live/testdata/kitchenSink-go/main.go` [here](https://github.com/blocky/delphi/blob/8196fb486008cd32a623312a3e2a455eefbe0af5/test/live/testdata/kitchenSink-go/main.go#L268). Might be worth using something of that nature if that appeals to you. E.g.:
```go
func writeErr(err string) uint64 {
...
}
writeOutput(output Output) uint64 {
...
}

func writeResultToSharedMem(result Result) uint64 {
	outputData, err := as.Marshal(result)
	if err != nil {
		// We panic on errors we cannot communicate back to the caller of this
		// function.
		panic("Fatal error: could not marshal output data")
	}
	return as.ShareWithHost(outputData)
}
```
this is not a request of changes, just wanted to point you towards the examples in delphi that have been getting cleaned up via iteration.

fetch_and_process_api_call/main.go
```diff
@@ -0,0 +1,136 @@
...
+func writeOutputToSharedMem(price Price, respErr error) uint64 {
+	isErr := false
+	errString := ""
+	if respErr != nil {
+		isErr = true
+		errString = "Error executing function: " + respErr.Error()
+	}
+
+	output := Output{Value: price, Error: errString, IsErr: isErr}
+	outputData, err := json.Marshal(output)
```
matteobjornsson, line 44: use `as.Marshal` here if you're going to keep marshal/unmarshal in `as`

matteobjornsson (approved):
Looks good. Works on a fresh EC2 instance, so the removal of the nix files with the C import seems to be fine. 

Some code organization comments, all up to you if you want to implement or not

fetch_and_process_api_call/main.go
matteobjornsson: also, if you are using `as.ShareWithHost` to leak to shared mem, might be worth updating the naming of the helper functions to be along the lines of `shareOutputWithHost` rather than `writeOutputToSharedMem` since the latter is somewhat referring to an underlying implementation in the `as` package that the users of the `as` package should be unaware of. 

If you were interested in the suggested changes above, the naming would be more like 
```go
func shareErr(err string) uint64 {
...
}
shareOutput(output Output) uint64 {
...
}

func shareResultWithHost(result Result) uint64 {
...
}
```
or something like that

fetch_and_process_api_call/main.go
mwittie: done

fetch_and_process_api_call/main.go
mwittie: OK, I made the changes. FWIW, I changed `shareErr` to `emitErr` and `shareOutput` to `emitPrice`. We can circle back to these conventions later as we finalize the SDK and the conventions.

fetch_and_process_api_call/main.go
```diff
@@ -0,0 +1,144 @@
...
+
+	return emitPrice(price)
+}
+
+func main() {}
+
+func emitErr(err error) uint64 {
+	result := Result{
+		Success: false,
+		Error:   err.Error(),
```
matteobjornsson, line 124: Take care this can panic without the err != nil check. It was one reason the signature was for a string not an error type in the examples I gave

fetch_and_process_api_call/main.go
matteobjornsson: Sounds good!

fetch_and_process_api_call/main.go
mwittie: ah, makes sense, I'll update

PR #3, opened 2025-02-16
Feature/bky 6325 iterate for lucid/examples

mwittie:
This PR updates the example of how to creates a TWAP using iterative calls to Blocky AS


iterate/Makefile
mwittie, line 1: Moving `iterate` to `twap` to name the example after the usecase, rather than the pattern. I think down the road we could have different examples using similar patterns. It also makes it easier to refer to the use case when talking with Lucid as twap, rather than iterate.

on-chain/contracts/User.sol
```diff
@@ -48,16 +48,16 @@ contract User is Ownable, TAParser {
...
+        uint twapIdx = 8;
+        string memory resultTWAP = JsmnSolLib.getBytes(out, tokens[twapIdx].start, tokens[twapIdx].end);
 
-        console.log("Did it Error:", isErr);
-        console.log("Average:", avg);
+        console.log("Success:", resultSuccess);
+        console.log("TWAP:", resultTWAP);
 
-        emit AttestedFunctionCallOutput(isErr, avg);
+        emit AttestedFunctionCallOutput(resultSuccess, resultTWAP);
```
mwittie, line 60: The changes here process the TWAP TA. As discussed, my plan is to connect with Lucid to see if this is not a problem for them. If its is, we can roll back to the Window TA

twap/Makefile
mwittie, line 1: Mostly copied from fetch_and_process_api_call example

twap/Makefile
```diff
@@ -0,0 +1,49 @@
...
+	@mkdir tmp
+
+check:
+	@command -v docker >/dev/null 2>&1 || { echo >&2 "docker not found in PATH. Aborting."; exit 1; }
+	@command -v bky-as >/dev/null 2>&1 || { echo >&2 "bky-as not found in PATH. Aborting."; exit 1; }
+	@command -v jq >/dev/null 2>&1 || { echo >&2 "jq not found in PATH. Aborting."; exit 1; }
+
+tmp/x.wasm: main.go $(wildcard as/*.go) $(wildcard price/*.go) | tmp check
+	@echo "Building WASM module..."
+	@docker run --rm \
```
mwittie, line 16: Switched from nix to docker build

twap/Makefile
```diff
@@ -0,0 +1,49 @@
...
+        tinygo build -o tmp/x.wasm -target=wasi main.go
+
+build: tmp/x.wasm
+
+init: build
+	@sed -e 's/VAR_EATTEST/null/' -e 's/VAR_TATTEST//' ./iteration-call.json.template > ./tmp/iteration-call.json
+	@cat tmp/iteration-call.json | ./bky-as attest-fn-call  | jq . > tmp/prev.json
+	@jq -r '.function_calls[0].claims.output | @base64d' tmp/prev.json | jq .
+
+iteration: build | check
```
mwittie, line 29: invoke iteration

twap/Makefile
```diff
@@ -0,0 +1,49 @@
...
+	@jq -r '.function_calls[0].logs | @base64d' tmp/prev.json
+
+twap: build | check
+	@sed \
+		-e 's/VAR_TATTEST/"tAttest": ${prev_ta},/' \
+		-e 's/VAR_EATTEST/${prev_ea}/' \
+		twap-call.json.template > tmp/twap-call.json
+	@cat tmp/twap-call.json | ./bky-as attest-fn-call | jq . > tmp/twap.json
+	@jq -r '.function_calls[0].claims.output | @base64d' tmp/twap.json | jq .
+	@jq -r '.function_calls[0].logs | @base64d' tmp/twap.json
```
mwittie, line 45: Added a separate target for `twap` to get a small attestation out based on data that's collected through `iterations`. Lucid wanted a smaller iteration to build on chain and I think it's cleaner than providing a window without samples, which creates a question of where twap value same from since the samples are blank.

twap/go.mod
```diff
@@ -0,0 +1,11 @@
...
+
+go 1.22
+
+require github.com/stretchr/testify v1.10.0
+
+require (
+	github.com/davecgh/go-spew v1.1.1 // indirect
+	github.com/pmezard/go-difflib v1.0.0 // indirect
+	gopkg.in/yaml.v3 v3.0.1 // indirect
+)
```
mwittie, line 11: I added these dependencies to make testing of the TWAP function easier. I verified through that adding them does not increase the size of `x.wasm`

twap/main.go
```diff
@@ -0,0 +1,219 @@
...
+
+type ResultPriceSamples struct {
+	Result       Result
+	PriceSamples PriceSamples
+}
+
+type ResultTWAP struct {
+	Result Result
+	TWAP   float64
+}
```
mwittie, line 41: I don't know if there's a smarter way of doing this in Go, but I wanted to have different output for the iterate and twap functions, while keeping the `Result` component the same

twap/main.go
```diff
@@ -0,0 +1,219 @@
...
+
+	return prevResult.PriceSamples, nil
+
+}
+
+func getNewPriceSample(tokenAddress string, chainID string) (price.Price, error) {
+	req := as.HostHTTPRequestInput{
+		Method: "GET",
+		URL: fmt.Sprintf(
+			"https://app.steer.finance/api/token/price?tokenAddress=%s&chainId=%s",
```
mwittie, line 91: Switched to API Lucid wanted to use

twap/main.go
```diff
@@ -0,0 +1,219 @@
...
+
+	now, err := as.TimeNow()
+	if err != nil {
+		return price.Price{}, fmt.Errorf("getting current time: %w", err)
+	}
+
+	return price.Price{
+		Value:     steerData.Price,
+		Timestamp: now,
+	}, nil
```
mwittie, line 122: Steer API does not provide a timestamp, so we create it ourselves

twap/main.go
```diff
@@ -0,0 +1,219 @@
...
+	nextPriceSamples := append(priceSamples, newPriceSample)
+	if len(nextPriceSamples) > args.NumSamples {
+		nextPriceSamples = nextPriceSamples[1:]
+	}
+
+	return emitPriceSamples(nextPriceSamples)
+}
+
+//export twap
+func twap(inputPtr, secretPtr uint64) uint64 {
```
mwittie, line 156: Added and exported the `twap` function. Its really nice we can export more than one function!

twap/price/price.go
```diff
@@ -0,0 +1,48 @@
...
+	"sort"
+	"time"
+)
+
+type Price struct {
+	Value     float64   `json:"price"`
+	Timestamp time.Time `json:"timestamp"`
+}
+
+func TWAP(samples []Price) (float64, error) {
```
mwittie, line 14: I created a TWAP utility function with its own testing. Initially I played with the idea that the TWAP would be calculated up to the current time (passed in as an additional parameter). I think that's the right way to do it, but it made the value of the TWAP change whenever I called this function, since it depended on the samples, but also the current time that would change the weight of the most recent sample.
To avoid the dynamic behavior, I changed the function to its current form, where we calculate the twap based on the samples alone. That makes the value of the least recent function not matter, as you see in the comment below.

mwittie: @dlm Could you pay particular attention to the implementation of this function in your review?

twap/Makefile
```diff
@@ -0,0 +1,52 @@
+prev_ta := $(shell jq '.function_calls[0].transitive_attestation' tmp/prev.json )
+prev_ea := $(shell jq '.enclave_attested_application_public_key.enclave_attestation' tmp/prev.json | sed 's/\//\\\//g' )
```
mwittie, line 2: When testing against the beta.3 server, I realized that the EA contains forward slashes that were messing with sed syntax during `VAR_EATTEST` replacements. To avoid that, I'm escaping the forward slashes here

twap/main.go
```diff
@@ -0,0 +1,219 @@
...
+) (
+	PriceSamples,
+	error,
+) {
+	// bootstrap with empty samples if we don't have a transitive attestation
+	if tAttest == nil {
+		return PriceSamples{}, nil
+	}
+
+	verifyOut, err := as.VerifyAttestation(
```
dlm, line 56: This is probably fine for now, but I think we may want to avoiding this pattern so high up and maybe improve the types a little in here so that it isn't just passing around byte arrays (at the very least, they could be named types that are actually byte arrays).

twap/main.go
```diff
@@ -0,0 +1,219 @@
...
+		return nil, fmt.Errorf("could not verify previous attestation: %w", err)
+	}
+
+	var fixedRep [][]byte
+	err = json.Unmarshal(verifyOut.RawClaims, &fixedRep)
+	if err != nil {
+		return nil, fmt.Errorf("could not unmarshal previous claims: %w", err)
+	}
+
+	prevResultData := fixedRep[3]
```
dlm, line 73: I wonder if we want to make these named constants.

twap/main.go
```diff
@@ -0,0 +1,219 @@
...
+	}
+
+	steerData := struct {
+		Price float64 `json:"price"`
+	}{}
+
+	err = json.Unmarshal(resp.Body, &steerData)
+	if err != nil {
+		return price.Price{}, fmt.Errorf(
+			"unmarshaling coin gecko data: %w...%s",
```
dlm, line 108: this error doesn't make sense if the data is coming from steer.

twap/main.go
```diff
@@ -0,0 +1,219 @@
...
+}
+
+//export iteration
+func iteration(inputPtr, secretPtr uint64) uint64 {
+	var args ArgsIterate
+	inputData := as.Bytes(inputPtr)
+	err := json.Unmarshal(inputData, &args)
+	if err != nil {
+		outErr := fmt.Errorf("could not unmarshal args args: %w", err)
+		return emitErr(outErr.Error())
```
dlm, line 132: FWIW, I am not a huge fan "emit" terminology.  It has a specific Prog Lang meaning (especially to those familiar with Node).  It made some sense in the first mode of computation that in the POC and before we came to some conclusion of the standard interface, but is kind of confusing when paired with the standard interface.  Specifically, you can "emit" things, but since we only sign and return the return value to the user, all the other stuff that was emitted goes to a black hole.

twap/main.go
```diff
@@ -0,0 +1,219 @@
...
+		outErr := fmt.Errorf("computing TWAP: %w", err)
+		return emitErr(outErr.Error())
+	}
+
+	return emitTWAP(twap)
+}
+
+func main() {}
+
+func emitErr(err string) uint64 {
```
dlm, line 182: see comment above about the "emit" naming of these functions.  I think it could be kind of confusing.

twap/price/price.go
```diff
@@ -0,0 +1,48 @@
...
+	if len(samples) == 0 {
+		return 0, fmt.Errorf("no samples provided")
+	}
+	if len(samples) == 1 {
+		return samples[0].Value, nil
+	}
+
+	// Sort samples from latest to earliest
+	sort.Slice(
```
dlm, line 23: FWIW, to make it clear what is going on, I would suggest updating this to:
```
lessThan := func(i, j int) bool { return samples[i].Timestamp.After(samples[j].Timestamp) }
sort.Slice(samples, lessThan)
```

twap/price/price.go
```diff
@@ -0,0 +1,48 @@
...
+		samples, func(i, j int) bool {
+			return samples[i].Timestamp.After(samples[j].Timestamp)
+		},
+	)
+
+	var weightedSum, totalWeight float64
+
+	// IMPORTANT: The value of the last sample is not included in the calculation
```
dlm, line 31: It shouldn't matter much since we don't know the true function that we are estimating but but since we are omitting the value of the last sample, we may want to use the midpoint of the range for the estimator.  That is, instead of using `prev.Value` on line 38, use `(prev.Value + next.Value) / 2`.

twap/Makefile
mwittie: Sounds good

twap/Makefile
```diff
@@ -0,0 +1,52 @@
+prev_ta := $(shell jq '.function_calls[0].transitive_attestation' tmp/prev.json )
```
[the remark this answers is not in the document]
mwittie: I double checked starting with `make clean`, which removes `tmp`. The readme flow works.

twap/Makefile
mwittie: I tried `-r` to make sure, but that's not the problem. The EA contains strings like `t1b/NhX4`. The forward slash messes with the `sed` replacement strings in the `iteration` and `twap` targets.

I'm going to move this forward slash replacement down into those targets, since that's where its needed, so that `prev_ea` is not modified here for a relatively far away reason.

twap/Makefile
```diff
@@ -0,0 +1,52 @@
...
+echo:
+	@echo "prev_ea: ${prev_ea}"
+
+iteration: build | check
+	@sed \
+		-e 's/VAR_TATTEST/"tAttest": ${prev_ta},/' \
+		-e 's/VAR_EATTEST/${prev_ea}/' \
+		iteration-call.json.template > tmp/iteration-call.json
+	@cat tmp/iteration-call.json | ./bky-as attest-fn-call | jq . > tmp/prev.json
+	@jq -r '.function_calls[0].claims.output | @base64d' tmp/prev.json | jq .
```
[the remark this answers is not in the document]
mwittie: Will update

twap/as/time.go
```diff
@@ -0,0 +1,52 @@
...
+
+import (
+	"encoding/json"
+	"fmt"
+	"time"
+)
+
+// TimeNow fetches the current UTC time from an external API.
+// In the future this will be implemented as a host function.
+func TimeNow() (time.Time, error) {
```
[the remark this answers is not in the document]
mwittie: That did it! I read a bit about these two formats. Good to know.

twap/main.go
mwittie: I agree that it would be nice to have a result type in the `as` library. I think this will come back around as we use the library for different examples, so I won't make a card for this yet.

on-chain/README.md
```diff
@@ -12,6 +12,8 @@ Use `nix` to set up a shell environment with the correct dependencies:
 nix-shell
 ```
 
+> Note that this process can take about 10 minutes the first time you run it.
```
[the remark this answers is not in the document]
mwittie: You're right. I'll remove.

twap/main.go
mwittie: I agree that this could be locked down a bit. I'll leave as is for now - we should address this when looking to improve `verify.go`

twap/main.go
mwittie: udpating

mwittie: updating

twap/main.go
mwittie: OK makes sense. I'll update to `write` following the terminology in https://github.com/blocky/delphi/blob/8196fb486008cd32a623312a3e2a455eefbe0af5/test/live/testdata/kitchenSink-go/main.go#L268.

twap/main.go
mwittie: updateing

twap/main.go
mwittie: OK, I added a named constant. We can do something smoother down the road

twap/price/price.go
mwittie: will update

twap/main.go
```diff
@@ -0,0 +1,219 @@
...
+	TAttest   json.RawMessage `json:"tAttest"`
+	Whitelist json.RawMessage `json:"whitelist"`
+}
+
+type Result struct {
+	Success bool
+	Error   string
+}
+
+type PriceSamples []price.Price
```
dlm, line 31: It is not clear to me why we are creating this type.  Seems unnecessary.

twap/as/verify.go
```diff
@@ -22,6 +22,14 @@ type HostVerifyAttestationOutput struct {
 	RawClaims []byte `json:"raw_claims"`
 }
 
+const (
+	ATTEST_FN_CALL_HASH_OF_CODE_IDX    = 0
```
dlm, line 26: In go, it is uncommon to use snake case for constants, just use cammel case with a capital if it is public or a lower case if it is private as you would do with any other constant.

For example: https://go.dev/doc/effective_go#parallel

twap/as/time.go
```diff
@@ -0,0 +1,38 @@
...
+	timeData := struct {
+		DateTime string `json:"dateTime"`
+	}{}
+
+	err = json.Unmarshal(resp.Body, &timeData)
+	if err != nil {
+		return time.Time{}, fmt.Errorf(
+			"unmarshaling time data: %w...%s",
+			err,
+			resp.Body,
```
dlm, line 30: heads up, this is a byte array but it is using a string in the format.  that is the same as casting the byte array as a string (as opposed to %v, which prints the byte array directly)  Not saying it is wrong or right, just making sure you are doing what you intend.

twap/as/time.go
```diff
@@ -0,0 +1,38 @@
...
+		)
+	}
+
+	// convert datetime from ISO 8601 format to RFC 3339 format then parse
+	parsedTime, err := time.Parse(time.RFC3339, timeData.DateTime+"Z")
```
dlm, line 35: this is not handling errors correctly.

dlm (approved):
Some bugs that should be fixed and a few comments that should be easy to address, but I think it all pretty simple to fix.

on-chain/contracts/User.sol
```diff
@@ -8,7 +8,7 @@ import {TAParser} from "./TAParser.sol";
 import {console} from "hardhat/console.sol";
 
 contract User is Ownable, TAParser {
-    event AttestedFunctionCallOutput(string result, string error);
+    event AttestedFunctionCallOutput(string Success, string TWAP);
```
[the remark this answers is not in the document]
mwittie: I wanted to swap it to reflect success, since Matteo wanted us to report success to avoid default false value in isErr.

I looked into this more and the convention is to report just the necessary information - in this case the attested output. If we get an error in the TA, I updated the transaction to revert, instead of writing to logs.

twap/price/price_test.go
```diff
@@ -0,0 +1,79 @@
...
+
+	tests := []struct {
+		name     string
+		samples  []Price
+		expected float64
+	}{
+		{
+			name: "multiple samples",
+			samples: []Price{
+				{Value: 100, Timestamp: now.Add(-1 * time.Hour)},
```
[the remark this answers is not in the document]
mwittie: adding

twap/main.go
```diff
@@ -0,0 +1,219 @@
...
+
+	newPriceSample, err := getNewPriceSample(args.TokenAddress, args.ChainID)
+	if err != nil {
+		outErr := fmt.Errorf("getting new sample: %w", err)
+		return emitErr(outErr.Error())
+	}
+
+	nextPriceSamples := append(priceSamples, newPriceSample)
+	if len(nextPriceSamples) > args.NumSamples {
+		nextPriceSamples = nextPriceSamples[1:]
```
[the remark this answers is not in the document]
mwittie: Good catch. I switched this algo to remove the least recent sample, but didn't think of this case.

twap/main.go
mwittie: OK

twap/as/verify.go
mwittie: ok

twap/as/time.go
mwittie: thanks

PR #6, opened 2025-02-28
[BKY-6341] hello world errors

mwittie:
This PR adds an example on how to handle errors in user functions

@dlm I took a stab at this, but let's circle up if you think there is a better way of implementing the pattern.


dlm (changes requested):
Lets circle up to discuss potentially putting some of the normal eng processes in place in this project.  I think there are some standard workflows, checklists, and safety rails that are not in place for this repository that are creating some unneeded frictions.

PR #7, opened 2025-03-01
[BKY-6378] Add error handling example

mwittie:
This PR adds an example on how to handle errors in user functions

- [x] this example includes a reference to hello-world on chain, so should be merged after https://github.com/blocky/attestation-service-examples/pull/5


error_handling/main.go
```diff
@@ -0,0 +1,55 @@
...
+	"github.com/blocky/as-demo/as"
+)
+
+type Result struct {
+	Success bool
+	Value   any
+}
+
+func (r Result) jsonMarshalWithError(err error) []byte {
+	resultStr := fmt.Sprintf(`{ "Success": false, "Value": "%v" }`, err)
```
mwittie, line 17: While `err` could be `nil`, the `Sprintf` does not panic and instead serialized `nil` error as "<nil>"

error_handling/README.md
```diff
@@ -0,0 +1,223 @@
...
+Before we get to error handling, let's create a function that successfully
+returns structured data. In [`main.go`](./main.go), we define the
+`successFunc` function:
+
+```go
+func successFunc(inputPtr, secretPtr uint64) uint64 {
+	type Output struct {
+		Number int `json:"number"`
+	}
+	output := Output{42}
```
dlm, line 59: Note that while I do sometimes do this in quick examples, it is often recommended to be explicit.  For example:

```
output := Output{ Number: 42 }
```
instead of 
```
output := Output{42}
```

error_handling/README.md
```diff
@@ -0,0 +1,223 @@
...
+
+```go
+func (r Result) jsonMarshalWithError(err error) []byte {
+	if err == nil {
+		err = errors.New("jsonMarshalWithError invoked with nil error")
+	}
+	resultStr := fmt.Sprintf(
+		`{ "Success": false, "Error": "%v" , "Value": null }`,
+		err,
+	)
```
dlm, line 157: I wonder if it would be better to use
```
resultStr := fmt.Sprintf(
    `{ "Success": false, "Error": "%s", "Value: null }`, 
    err.Error(),
)
```

The %v verb is go specific as where %s is used in most Sprintf type of environments, so more developers will know it already.

error_handling/README.md
```diff
@@ -0,0 +1,223 @@
...
+	data := []byte(resultStr)
```
dlm, line 158: FWIW, you could update to 
```
return []byte(resultStr)
```
and remove a line and an extra variable.

error_handling/main.go
```diff
@@ -0,0 +1,62 @@
...
+	Success bool
+	Error   string
+	Value   any
+}
+
+func (r Result) jsonMarshalWithError(err error) []byte {
+	if err == nil {
+		err = errors.New("jsonMarshalWithError invoked with nil error")
+	}
+	resultStr := fmt.Sprintf(
```
dlm, line 21: I made some comments about this code up in the readme

dlm (approved):
Looks good.  A few very minor comments.  :shipit:

error_handling/README.md
mwittie: Updating

error_handling/README.md
mwittie: Sure, good suggestion. Now that we are safe to use err.Error it will be cleaner

error_handling/README.md
mwittie: updating

PR #8, opened 2025-03-04
Feature/bky 6379 update fetch api/example

mwittie:
This PR updates the api fetch example with a description.


error_handling/README.md
mwittie, line 1: The changes in error handling basically just move error handling into output.go

coin_prices_from_coingecko/output.go
mwittie, line 1: Here's the recycled output.go for the coin prices example

coin_prices_from_coingecko/README.md
mwittie, line 1: This is where the bulk of the changes are. I took the description from what we had on the docs and modified it to work within the examples repository

twap/config.toml
mwittie, line 1: Adding a missing config file

coin_prices_from_coingecko/README.md
```diff
@@ -0,0 +1,283 @@
...
+	if err != nil {
+		outErr := fmt.Errorf("getting price: %w", err)
+		return WriteError(outErr)
+	}
+
+	return WriteOutput(price)
+}
+```
+
+The function takes two `uint64` arguments and returns a `uint64`. These are fat
```
dlm, line 143: Consider adding a tutorial for processing input and secrets and instead of explaining it here, explain it in the input and secret input tutorial and just reference it in this one.

coin_prices_from_coingecko/output.go
```diff
@@ -0,0 +1,44 @@
...
+	"github.com/blocky/as-demo/as"
+)
+
+type Result struct {
+	Success bool
+	Error   string
+	Value   any
+}
+
+func (r Result) jsonMarshalWithError(err error) []byte {
```
dlm, line 17: If we are going with a design in which all functions are expolrted, consider exporting this function as well.

error_handling/output.go
```diff
@@ -0,0 +1,44 @@
...
+	"github.com/blocky/as-demo/as"
+)
+
+type Result struct {
+	Success bool
+	Error   string
+	Value   any
+}
+
+func (r Result) jsonMarshalWithError(err error) []byte {
```
dlm, line 17: similar comment as before.

dlm (approved):
Looks good.  Only real comment is that I think we should add another example that focuses on input and secret input, and then reference it in this one.  I don't think it is a blocker for merging, but it could help improve one of the rougher edges in the system.

coin_prices_from_coingecko/README.md
mwittie: I thought about it, but thought that it would be a bit too in a vacuum. We can certainly circle back to this idea though.

For now, I'm going to add a directory bullets to to the top level README, so that that people can at least find where to learn about input and output

coin_prices_from_coingecko/output.go
mwittie: updating

error_handling/output.go
mwittie: updating

PR #13, opened 2025-03-18
[BKY-6444] Refactor TA Parser Lib

dlm:
In this PR, Dave re-wrote the TA Parser and user smart contracts so that:
1. there is only 1 tx needed for processing a TA
2. TAParserLib follows the solidity style guide
3. TAParserLib more clearly defines what is TA specific and what is specific
   to a TA for a function call.

Bear then updated `test/user.ts` to address the TypeScript linter warnings and errors. Follow-up cards were created to apply the changes in this PR to the rest of our "on-chain" related examples as well as to add linting for solidity and typescript to the Makefile


hello_world_on_chain/contracts/User.sol
```diff
@@ -2,33 +2,20 @@
 pragma solidity ^0.8.10;
 
 import {TAParserLib} from "../lib/TAParserLib.sol";
-import {Ownable} from "@openzeppelin/contracts/access/Ownable.sol";
```
dlm, line 5: Example contract no longer needs to be ownable.  If we want to make a more complex user contract, we can still do that as a more "advanced usage section"

hello_world_on_chain/contracts/User.sol
```diff
@@ -2,33 +2,20 @@
...
-    function verifyAttestedFnCallClaims(
-        string calldata taData
+    function demo(
+        bytes calldata publicKey,
+        string calldata transitiveAttestation
     )
-    public
+        public
     {
-        TAParserLib.FnCallClaims memory claims = TAParserLib.parseTA(
```
dlm, line 29: To get us ready for moving to a library that we make available though NPM, I added some updates to the TAParserLib.  Previous iteration of the contract assumed that All TAs are fun call claims, this change allows for more diversity of TA parsing.

hello_world_on_chain/lib/TAParserLib.sol
```diff
@@ -20,10 +20,22 @@ library TAParserLib {
         string Output;
     }
 
+    function verifyAttestedFnCall(
```
dlm, line 23: This interface seems, to me, to be a point of friction.  We would like to make it easy for a user to bring data on chain and so we should take in our key and then convert it, however, we should also make it easy for a user to follow the better pattern of not converting the public key to an address every time.  I am temped to have this function take in an address and make the user contract have 2 lines instead of 1 and take in an address not the bytes of the public key thoughts?

hello_world_on_chain/lib/TAParserLib.sol
```diff
@@ -20,10 +20,22 @@ library TAParserLib {
...
+        string memory taData = verifyTA(publicKeyAsAddress, transitiveAttestation);
+        TAParserLib.FnCallClaims memory claims = decodeFnCallClaims(taData);
+        return claims;
+    }
+
     function publicKeyToAddress(
         bytes calldata publicKey
     )
-    internal pure returns (address)
+        internal pure returns (address)
```
dlm, line 38: while working on this contract, I found the solidity smart contract style guide https://docs.soliditylang.org/en/latest/style-guide.html and so I updated (most of the code to follow it) thoughts?

hello_world_on_chain/lib/TAParserLib.sol
```diff
@@ -34,11 +46,11 @@ library TAParserLib {
         return address(uint160(uint256(keccak256(strippedPublicKey))));
     }
 
-    function parseTA(
-        string calldata taData,
-        address publicKeyAddress
+    function verifyTA(
```
dlm, line 49: Biggest changes that I made to the structure of this is that the verify of the TA and the parsing of the data in the TA to FN call claims are decoupled.  That allows a user to more clearly see layers of the encoding and verification (and could allow us to easily add other transitively attested things)

hello_world_on_chain/lib/TAParserLib.sol
```diff
@@ -78,6 +80,10 @@ library TAParserLib {
         uint success;
         (success, tokens, number) = JsmnSolLib.parse(taData, 3);
 
+        // The flat rep of a TA is an array of 2 byte arrays.
+        // As such, we expect to have 3 tokens.
+        // Item 0 is the array of byte arrays with 1 and 2 as the content.
+        require(number == 3, "Wrong number of tokens");
```
dlm, line 86: Added some checks in here to make sure that data is correctly formatted.

hello_world_on_chain/lib/TAParserLib.sol
```diff
@@ -96,8 +102,12 @@ library TAParserLib {
         JsmnSolLib.Token[] memory tokens;
         uint number;
         uint success;
-        (success, tokens, number) = JsmnSolLib.parse(b64, 20);
+        (success, tokens, number) = JsmnSolLib.parse(b64, 6);
```
dlm, line 105: more data checking.

hello_world_on_chain/lib/TAParserLib.sol
```diff
@@ -106,4 +116,13 @@ library TAParserLib {
 
         return claims;
     }
+
+    function base64d(
```
dlm, line 120: updated some of the order of the functions to conform the the solidity style guide.

hello_world_on_chain/test/user.ts
```diff
@@ -69,11 +69,10 @@ describe("Local Test", function () {
         const publicKey = evmLinkData.publicKey;
 
         const {userContract} = await loadFixture(deployUser);
-        await userContract.setTASigningKeyAddress(publicKey as any);
 
         // when
         const ta = evmLinkData.transitiveAttestation;
-        const tx = await userContract.verifyAttestedFnCallClaims(ta as any)
+        const tx = await userContract.demo(publicKey, ta)
```
dlm, line 75: Updated the name of the function that the user calls.  I thought it made it a little more clear what is going on, alternative names welcome.

hello_world_on_chain/package.json
```diff
@@ -9,7 +9,6 @@
   "license": "MIT",
   "dependencies": {
     "@nomicfoundation/hardhat-toolbox": "^5.0.0",
-    "@openzeppelin/contracts": "^5.2.0",
```
dlm, line 12: since we simplified, we no longer need the user contract to be ownable and so we don't need openzeppelin

hello_world_on_chain/test/user.ts
```diff
@@ -69,11 +69,10 @@ describe("Local Test", function () {
         const publicKey = evmLinkData.publicKey;
 
         const {userContract} = await loadFixture(deployUser);
-        await userContract.setTASigningKeyAddress(publicKey as any);
 
         // when
         const ta = evmLinkData.transitiveAttestation;
-        const tx = await userContract.verifyAttestedFnCallClaims(ta as any)
```
dlm, line 76: There were some spots with any casts others without it.  Was there any reason that they were in there?  From playing around with this, it seemed like the casts were not needed.

hello_world_on_chain/lib/TAParserLib.sol
```diff
@@ -34,11 +46,11 @@ library TAParserLib {
         return address(uint160(uint256(keccak256(strippedPublicKey))));
     }
 
-    function parseTA(
-        string calldata taData,
-        address publicKeyAddress
+    function verifyTA(
+        address publicKeyAddress,
+        string calldata taData
```
mwittie, line 51: I think `taData` is confusing here since we also have TA.Data. Maybe we could rename this variable to `transitiveAttestation` or `taBytes`?

hello_world_on_chain/contracts/User.sol
```diff
@@ -2,33 +2,20 @@
...
-        bytes calldata taSigningKey
-    )
-    public onlyOwner
-    {
-        taSigningKeyAddress = TAParserLib.publicKeyToAddress(taSigningKey);
-    }
-
-    function verifyAttestedFnCallClaims(
-        string calldata taData
+    function demo(
```
mwittie, line 10: I think `demo` lacks context here to someone just jumping in to read this function. Maybe we could name it something like `processTAHelloWorld` since were processing a transitively attested call to `helloWorld`?

hello_world_on_chain/test/user.ts
mwittie: My IDE complains here if we don't have the `as any`
image

From what I can tell, I can reduce the level of warning in the IDE on Type Mismatch, but the right thing is to add `as any`

mwittie: In the Sepolia test below the complain is not about `as any` (which is maybe why I didn't have it there), but about the `demo` function
image

I looked into getting my IDE to see the symbol and found a solution to be 

image

where I declare the interface and the load the contract as the interface. Interestingly, I don't get complaints about parameter type here.

mwittie (commented):
Looks pretty good. Couple suggestions, but I think we can move ahead

hello_world_on_chain/contracts/User.sol
dlm: done

hello_world_on_chain/contracts/User.sol
```diff
@@ -2,33 +2,20 @@
...
+        bytes calldata publicKey,
```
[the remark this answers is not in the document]
dlm: done

hello_world_on_chain/contracts/User.sol
dlm: updated docs.

hello_world_on_chain/test/user.ts
```diff
@@ -57,23 +57,29 @@ const loadUserContractABI = () => {
     }
 }
 
+interface UserContract extends ethers.Contract {
```
dlm, line 60: @mwittie can you see if this update addresses the warnings that you are getting on your system in the typescript files and if not, can you fix?  I am not sure what tools your Ide is using that is generating the notifications and so I can't reproduce.  Out of scope for this PR, but, it could be worth making some decisions on how do you want to handle things like linting and code checks for this repo. Note that in https://blocky.atlassian.net/browse/BKY-6457 I am just setting up workflows and testing.

hello_world_on_chain/test/user.ts
dlm: Can you also check if this addresses the `any` suggestion.

Also out of scope, but I am kind of wondering if some of these issues are with how the tooling is configured.  Seems kind of weird that a correctly configured project would product "Unresolved function" errors yet still work.  Do you think it is worth looking in to?

hello_world_on_chain/README.md
```diff
@@ -243,7 +217,7 @@ which show that the [`User`](contracts/User.sol) contract was able to
 verify the TA in just over 2 seconds.
 If you go to Basescan to see 
 [contract transaction logs](https://sepolia.basescan.org/tx/0x82eb1d2cd500a1e236678a04f9ee8361e2fda11224d0791612ae2244d294bf78#eventlog)
-you can see that the `verifyAttestedFnCallClaims` emitted the
+you can see that the `processTAHelloWorld` emitted the
```
dlm, line 220: @mwittie since I don't have creds to run this, I did not yet update "Step 

Some updates that are definitely needed (there may be others that cascade):
1. some of the outputs have changed.
3. the contract has changed so the contract needs to be redeployed.
4. the value in `deployments/user_deployed_address` needs to be updated.
5. the link in the block explorer needs to be updated.

dlm:
@mwittie I made the updates to the code and the walkthough.  The next steps would be some back and forth on the warnings that I cannot reproduce and some figuring out creds and keys for Base.  Before I continue, since the plan is to hand off this task at some point, I was wondering if the current state would be a good point.

hello_world_on_chain/test/user.ts
```diff
@@ -1,79 +1,92 @@
...
 import hre from "hardhat";
 import {loadFixture} from "@nomicfoundation/hardhat-toolbox/network-helpers";
 import {expect} from "chai";
 import {ethers} from "ethers";
+import fs from "fs";
+import path from "path";
+import {User} from "../typechain-types";
 
-const fs = require("fs")
-const path = require("path")
```
tahardi, line 7: Recommended to use `import` over `require`

hello_world_on_chain/test/user.ts
```diff
@@ -1,79 +1,92 @@
+// noinspection DuplicatedCode
```
tahardi, line 1: Goland keeps giving me warnings about "duplicated code" highlighting that portions of this test file is the same in our other example folders. it's super annoying so I added this to turn it off

hello_world_on_chain/test/user.ts
```diff
@@ -1,79 +1,92 @@
...
+    transitiveAttestation: string;
+};
 
-function loadEVMLinkData(jsonPath: string) {
+function loadEVMLinkData(jsonPath: string): EVMLinkData {
     try {
-        const dir = path.resolve( __dirname, jsonPath);
-        const file = fs.readFileSync(dir);
+        const dir: string = path.resolve( __dirname, jsonPath);
+        const file: string = fs.readFileSync(dir, "utf8");
```
tahardi, line 19: I had to add the `"utf8"` argument after switching `fs` from `require` to `import`. Basically telling it to read in as a string because for whatever reason when we use `import` it defaults to bytes. Or it might be that it always defaulted to bytes but TypeScript couldn't perform analysis when `require` was being used

hello_world_on_chain/test/user.ts
```diff
@@ -1,79 +1,92 @@
...
 
-        const data = JSON.parse(file);
+        const data: any = JSON.parse(file);
```
tahardi, line 21: I'll only write this once: you'll notice that I've added explicit types to all variables and function definitions. I wasn't getting warnings per-se, but Goland was adding "ghost" script everywhere telling me what all the inferred types were. I found that more annoying than explicitly having the types written out in code and figured it would be helpful to those whose IDE does not show them inferred types. Plus, it's called "type" script so we might as well write them, right?

hello_world_on_chain/test/user.ts
```diff
@@ -1,79 +1,92 @@
...
+        const ta: string = Buffer.from(taBytes).toString('utf-8');
 
-        return {
-            publicKey: `0x${publicKeyHex}`,
-            transitiveAttestation: ta
-        };
+        return { publicKey: `0x${publicKeyHex}`, transitiveAttestation: ta};
     } catch (e) {
-        console.log(`e`, e)
+        throw new Error(`Error loading EVM link data: ` + e);
```
tahardi, line 35: I've made all of these helper load functions throw an error, whereas previously they simply logged to the console and _implicitly_ returned an `undefined` value. This implicit value was causing issues further down in the test where we would get warnings about using the outputs of these functions because they were potentially undefined.

Now that all of their returns are explicit and they throw errors we do not get warnings.

hello_world_on_chain/test/user.ts
```diff
@@ -1,79 +1,92 @@
...
     }
 }
 
-const loadUserDeployedAddress = () => {
+const loadUserDeployedAddress: () => string  = () : string  => {
```
tahardi, line 39: Again, explicitly return type or throw error. Previously the implied return was `string | undefined` which threw errors below when passing into another function that only took `string` and didn't appreciate the potential for an `undefined` value

hello_world_on_chain/test/user.ts
```diff
@@ -1,79 +1,93 @@
...
+        throw new Error(`loading user contract ABI: ` + e);
     }
 }
 
-describe("Local Test", function () {
-    async function deployUser() {
-        const contract = await hre.ethers.deployContract("User");
+interface UserContract extends ethers.Contract {
+    // @ts-ignore
+    processTAHelloWorld(publicKey: any, ta: any): Promise<ethers.ContractTransactionResponse>;
```
tahardi, line 69: This is the only Goland error I've been unable to address. I've tried defining methods in a number of different ways but it does not seem to like any of them. So, I've added an "ignore" directive, but if you want I can spend more time trying to fix it.

tahardi: Specifically, it gives the following error:
```
TS2411: Property processTAHelloWorld of type
(publicKey: any, ta: any) => Promise<ContractTransactionResponse>
is not assignable to string index type BaseContractMethod<any[], any, any
```

dlm (commented):
I cannot approve since I originally created it.  Looks fine to me, but there are still some items in here that would need to be rerun by @mwittie, specifically, some of the parts of this example that links to sepolia need to be updated.  I'll ping him again to remind him that it is still blocked on those steps.

hello_world_on_chain/contracts/User.sol
```diff
@@ -2,33 +2,23 @@
...
     event AttestedFunctionCallOutput(string output);
 
-    address private taSigningKeyAddress;
-
-    constructor() Ownable(msg.sender) {
-    }
-
-    function setTASigningKeyAddress(
-        bytes calldata taSigningKey
+    function processTAHelloWorld(
```
mwittie, line 10:
```suggestion
    function processTransitiveAttestedHelloWorld(
```

hello_world_on_chain/lib/TAParserLib.sol
```diff
@@ -34,13 +34,29 @@ library TAParserLib {
         return address(uint160(uint256(keccak256(strippedPublicKey))));
     }
 
-    function parseTA(
-        string calldata taData,
-        address publicKeyAddress
+    function verifyAttestedFnCall(
```
mwittie, line 37: The naming of the smart contract functions is a bit inconsistent.
processTAHelloWorld vs verifyAttestedFnCall - maybe <verb>TransitiveAttested<thing>?
verifyTA - maybe verifyTransitiveAttestation, that way we don't have a collision on TA being transitive attestation and trasitively attested

hello_world_on_chain/test/user.ts
mwittie: Is there a reason why these functions are declared as const, while loadEVMLinkData is declared as function?

hello_world_on_chain/test/user.ts
mwittie: confirmed I'm not getting IDE warnings

hello_world_on_chain/test/user.ts
```diff
@@ -93,22 +107,22 @@ describe("Base Sepolia Tests", function () {
...
+    const evmLinkData: EVMLinkData = loadEVMLinkData("../inputs/out.json");
 
-    it("Verify TA", async () => {
-        const ta = evmLinkData.transitiveAttestation;
-        const tx = await userContract.verifyAttestedFnCallClaims(ta)
+    it("Verify TA", async (): Promise<void> => {
+        const publicKey: string = evmLinkData.publicKey;
+        const ta: string = evmLinkData.transitiveAttestation;
+        const tx: ethers.ContractTransactionResponse = await userContract.processTAHelloWorld(
+            publicKey as any,
```
mwittie, line 118: I tied removing these `as any` and I'm no longer seeing errors. Can you check that you can do that on your end as well?

hello_world_on_chain/test/user.ts
```diff
@@ -93,22 +107,22 @@ describe("Base Sepolia Tests", function () {
...
+            ta as any
+        )
         // poll instead of tx.wait() to get the lowest possible delay
         for (; ;) {
-            const txReceipt = await provider.getTransactionReceipt(tx.hash);
+            const txReceipt: ethers.TransactionReceipt | null = await provider.getTransactionReceipt(
```
mwittie, line 123: This line is long. Possible to wrap?

mwittie (approved):
Looks good!

hello_world_on_chain/README.md
mwittie: udpated

hello_world_on_chain/contracts/User.sol
```diff
@@ -2,35 +2,28 @@
...
-        bytes calldata taSigningKey
-    )
-    public onlyOwner
-    {
-        taSigningKeyAddress = TAParserLib.publicKeyToAddress(taSigningKey);
-    }
-
-    function verifyAttestedFnCallClaims(
-        string calldata taData
+    function processTransitivelyAttestedHelloWorldOutput(
```
mwittie, line 10: I unified the naming of the functions here and in TAParser lib. See if that makes sense

hello_world_on_chain/test/user.ts
```diff
@@ -1,79 +1,95 @@
...
+    it("Verify TA", async (): Promise<void> => {
         // given
-        const evmLinkData = loadEVMLinkData("../inputs/out.json");
-        const publicKey = evmLinkData.publicKey;
+        const evmLinkData: EVMLinkData = loadEVMLinkData("../inputs/out.json");
+        const publicKey: string = evmLinkData.publicKey;
 
-        const {userContract} = await loadFixture(deployUser);
-        await userContract.setTASigningKeyAddress(publicKey as any);
+        const {userContract} = await loadFixture(deployUser) as UserContract;
```
mwittie, line 84: I added the `as UserContract` so that we can use the defined interface and don't have to add `as any` on line 121

hello_world_on_chain/test/user.ts
mwittie: The changes made by Bear and my addition to to load contract `as UserContract` addresses these issues

PR #19, opened 2025-03-20
[BKY-6457] Setup an environment

dlm:
This is in no way perfect, however it is intended to be part of the scaffolding to get some of the workflows in place.  We may want to take it out eventually or improve it, but in the meantime....

When preparing examples that  work in with specific versions of the AS service, it can be somewhat challenging 
to move between versions.  This PR adds a nix shell that will build a version of bky-as.  That way, when we work on 
beta4 it can set up an environment for beta 4 and sim for beta 5 (and for the next few days for beta5 pre releases.)   

Note that you will not be able to create a nix shell unless you have a PAT with access to various internal blocky repos.


matteobjornsson:
Would it make sense to add a very brief CONTRIBUTING.md or something similar? Normally for something like this setup/commands would go into a readme.

shell.nix
```diff
@@ -0,0 +1,54 @@
...
+
+  };
+in
+pkgs.mkShellNoCC {
+  packages = [
+    pkgs.git
+    pkgs.go
+    pkgs.gotools # for tools like goimports
+    pkgs.golangci-lint # for linting go files
+    pkgs.tinygo # for building wasm
```
matteobjornsson, line 47: do we need to pin `go` or `tinygo` to a version?

mwittie (approved):
Slick!

shell.nix
```diff
@@ -0,0 +1,54 @@
+{
+  githubPAT ? builtins.getEnv "GITHUB_PAT",
```
matteobjornsson, line 2: I was expecting this to work. I was assuming the GITHUB_PAT was supposed to result in skipping this password step? Am I approaching this wrong? `gh auth token` works outside of this command. 
Screenshot 2025-03-20 at 6 31 18 PM

shell.nix
dlm: the versions that is pinned by the version that was part of nixpkgs 24.11.  Which is go 1.23.6 and tinygo 0.34.   If we want to spend a few more cycles on it we may be able to tighten up those versions, but in the short term, my main goal is to figure out how to get versions of bky-as to aligh with the project.

PR #20, opened 2025-03-21
[BKY-6457] Add workflows and pr templates

dlm:
In this PR, we introduce some 
of our basic workflows a pull request template


dlm:
Merging without review so that I can get a few more tasks setup

PR #27, opened 2025-03-25
[BKY-6376] Random example

tahardi:
## Describe your changes
This PR adds an example Blocky AS guest function that demonstrates how to get random numbers from the Blocky AS runtime.

## Related Jira cards
[random example](https://blocky.atlassian.net/browse/BKY-6376?atlOrigin=eyJpIjoiYWE2M2UxNzM3YWViNDdhMTk0NjIwY2UzN2I5NzJlNzAiLCJwIjoiaiJ9)

- [x] This PR is small. 
    - Otherwise, justify here:


random/main.go
```diff
@@ -0,0 +1,45 @@
...
+	case err != nil:
+		outErr := fmt.Errorf("could not unmarshal input args: %w", err)
+		return WriteError(outErr)
+	case input.Sides == 0:
+		outErr := fmt.Errorf("die cannot have zero sides")
+		return WriteError(outErr)
+	}
+
+	randBytes := make([]byte, 8)
+	n, err := crand.Read(randBytes)
```
tahardi, line 31: Currently, we show that you can generate random numbers in much the same way you would a normal program. Namely, you import a standard library that pulls entropy from the system which it then uses to generate random streams. This works because tinygo provides a `crypto/rand` implementation which pulls entropy from the `wazero` runtime which we have correctly seeded.

If users don't want to (or can't) depend on having a stdlib implementation like `crypo/rand` we could define a host func in delphi that provides random byte streams by pulling from the `newman` entropy provider. I figure that is something we could push off for now and implement in the future if we see a need for it

random/README.md
```diff
@@ -0,0 +1,182 @@
...
+  32 bits are a memory address and the second 32 bits are the size of the data.
+  The memory space is sandboxed and shared between the TEE host program (Blocky
+  AS server) and the WASM runtime (your function). The `inputPtr` and
+  `secretPtr` arguments carry user-defined function input and secrets,
+  though we only make use of input in this example. The output of the function
+  is also a memory pointer, whose value will be returned to the user.
+- The function uses the `basm`
+  [Blocky Attestation Service WASM Go SDK](https://github.com/blocky/basm-go-sdk)
+  `basm.ReadFromHost` to fetch the input data and unmarshal it into the
+`Args` struct.
```
dlm, line 124: fix spacing.

random/README.md
```diff
@@ -0,0 +1,182 @@
...
+
+Finally, note that we are able to generate random numbers the way you would in
```
dlm, line 126: Why are some of the items in a bulleted list but this one is not?  ... do we do that in others?  From a random sample of a few (e.g. the hello world one, we do not).

random/README.md
```diff
@@ -0,0 +1,182 @@
...
+a normal Go program (i.e., via `crypto/rand`). This is because
+`tinygo` provides an implementation for `crypto/rand` and we properly seed the
```
dlm, line 128: If you were familiar with go, you may wonder why we are using crypto/rand to generate a random dice roll and not `rand.Intn`.  Is it worth a sentence?

random/main.go
dlm: I think it is fine.  Let's wait to see if anybody wants such a feature.

dlm (approved):
From my perspective, I think this is good.  Since this is your first PR to the project, please wait for a second review though.

random/README.md
tahardi: fixed

random/README.md
tahardi: Probably because this is some amalgamation of readmes from multiple examples. Either way, I slightly edited these last couple of sentences and made them a bullet as well

random/README.md
tahardi: Ahhh... Honestly, I was only using `crypto/rand` because I'd been using it recently for things that actually required cryptographically secure randomness.

I've updated the example to simply use `math/rand.Intn()` since emulating a die roll does not necessarily require `crypto/rand`. I've also updated these lines so that a reader is aware they could use either `math/rand` or `crypto/rand` depending on their use case

random/go.mod
```diff
@@ -0,0 +1,10 @@
+module github.com/blocky/as-demo
+
+go 1.22.6
+
+require github.com/blocky/basm-go-sdk v0.1.0-beta.4
```
mwittie, line 5: revisit if we're ready to update this dependency and maybe the on in shell.nix as well

random/README.md
```diff
@@ -0,0 +1,181 @@
...
+  "Error": "",
+  "Value": 3
+}
+```
+
+## Walkthrough
+
+### Step 1: Create a parameterized random function
+
+We'll demonstrate random by emulating a die roll in `rollDie` in
```
mwittie, line 53:
```suggestion
We'll demonstrate random number generation by emulating a die roll in `rollDie` in
```

random/README.md
```diff
@@ -0,0 +1,181 @@
...
+{
+  "code_file": "tmp/x.wasm",
+  "function": "rollDie",
+  "input": {
+    "sides": 20
+  }
+}
+```
+
+Notice the `input` section, which contains the parameters for `rollDie`,
```
mwittie, line 68: Just one parameter

random/README.md
```diff
@@ -0,0 +1,181 @@
...
+specifically the `sides` field set to `20`. Meaning, `rollDie` will return
+a number in the range [1,20].
+
+Next, we define the `rollDie` function in [`main.go`](./main.go):
+
+```go
+type Args struct {
+    Sides uint64 `json:"sides"`
```
mwittie, line 76: maybe `NumSides` or `DieSides`?

random/README.md
```diff
@@ -0,0 +1,181 @@
...
+//export rollDie
+func rollDie(inputPtr uint64, secretPtr uint64) uint64 {
+    var input Args
+    inputData := basm.ReadFromHost(inputPtr)
+    err := json.Unmarshal(inputData, &input)
+    switch {
+    case err != nil:
+        outErr := fmt.Errorf("could not unmarshal input args: %w", err)
+        return WriteError(outErr)
+    case input.Sides == 0:
```
mwittie, line 88: What if the is negative?

random/README.md
```diff
@@ -0,0 +1,181 @@
...
+        outErr := fmt.Errorf("die cannot have zero sides")
+        return WriteError(outErr)
+    }
+
+    randBytes := make([]byte, 8)
```
mwittie, line 93: What if the number of sides does not fit in 8 bytes?

random/README.md
```diff
@@ -0,0 +1,181 @@
...
+    }
+
+    roll := (binary.BigEndian.Uint64(randBytes) % input.Sides) + 1
+    return WriteOutput(roll)
+}
+```
+
+You will notice a few things:
+
+- The `rollDie` function is exported so that it can be invoked by the
```
mwittie, line 111: I think we can take this out - already covered by the hello-world example

random/README.md
```diff
@@ -0,0 +1,181 @@
...
+  Blocky AS server in a WASM runtime.
+- The function takes two `uint64` arguments and returns a `uint64`. These are
```
mwittie, line 113: Marcin has a card to remove the discussion of parameter parsing from the different examples. Since both of these could be in flight at the same time, you guys might want to coordinate as to who is removing this part from this example.

random/README.md
```diff
@@ -0,0 +1,181 @@
...
+
+Next, we define the `rollDie` function in [`main.go`](./main.go):
+
+```go
+type Args struct {
+    Sides uint64 `json:"sides"`
+}
+
+//export rollDie
+func rollDie(inputPtr uint64, secretPtr uint64) uint64 {
```
mwittie, line 80: I think this is out of date with respect to main.go. You might be able to ignore the next couple of comments.

random/README.md
```diff
@@ -0,0 +1,181 @@
...
+  though we only make use of input in this example. The output of the function
+  is also a memory pointer, whose value will be returned to the user.
+- The function uses the `basm`
+  [Blocky Attestation Service WASM Go SDK](https://github.com/blocky/basm-go-sdk)
+  `basm.ReadFromHost` to fetch the input data and unmarshal it into the
+  `Args` struct.
+- We are able to generate random numbers the way you would in a normal Go
+  program (i.e., via `math/rand` or `crypto/rand`). This is because `tinygo`
+  provides implementations for `math/rand` and `crypto/rand`, and because we
+  properly seed the `wazero` runtime with a source of entropy (`wazero`
```
mwittie, line 128: The major value of this example is that people can trust the random number is correct. I think that just saying we set it properly is not enough here. Could you go into the details of what we did?

random/fn-call.json
```diff
@@ -0,0 +1,7 @@
+{
+  "code_file": "tmp/x.wasm",
+  "function": "rollDie",
+  "input": {
+    "sides": 20
```
mwittie, line 5: `die_sides?

random/README.md
```diff
@@ -0,0 +1,181 @@
...
+  "Success": true,
+  "Error": "",
+  "Value": 3
+}
+```
+
+You will most likely see a different number for the `Value` field as it is
+randomly generated. On that note, subsequent invocations of `rollDie` should
+produce different numbers. You can control the range of said numbers by
+updating the `sides` field in the `fn-call.json` file.
```
mwittie, line 175: `die_sides`?

random/README.md
```diff
@@ -0,0 +1,181 @@
...
+
+## Next steps
+
+Now that you have successfully run the example, you can start modifying it to
+fit your own needs. Check out other examples in this repository, to learn what
```
mwittie, line 180: One thing you could ask people to do is extend the example with a message that says "Critical Failure" or "Critical Success". You could also suggest that they roll multiple different sided dies in the same call, since that's often what's needed in D&G (though I think D&G maybe trademarked, so we may want to be somewhat generic about that suggestion).

mwittie (approved):
Looks good. Just a few suggestions

random/.gitignore
```diff
@@ -0,0 +1 @@
+tmp/
```
dlm, line 1: this is a test

dlm: .... yep, that did .... if all of your comments become outdated, did your review ever exist?

random/fn-call.json
tahardi: updated to `die_sides` here, in `main.go`, and README

random/README.md
tahardi: done

random/README.md
tahardi: Ah good catch. Forgot to update the code here after changing in `main.go`. Fixed.

random/README.md
tahardi: Added a line stating that we seed wazero with entropy obtained from the Nitro Secure Module

random/README.md
tahardi: changed to DieSides

random/README.md
tahardi: Removed

PR #35, opened 2025-04-05
[BKY-6561] Add templated url

dlm:
## Describe your changes

In this PR, we add functionality to update the urls via templates.

The way that we make this work is by replacing all of the version tags in the urls of the examples with a template string
and then in the "stable shell" we introduce a function called `render-md` that renders all of the markdown files.

That way, we don't need to keep manually updating all the the urls for every release.

## Related Jira cards
[bky-6561](https://blocky.atlassian.net/browse/BKY-6561)

## Notes for reviewers

Let me know what y'all think.

- [x] This PR is small. 
    - Otherwise, justify here:


nix/mkDevShell.nix
```diff
@@ -22,6 +22,14 @@ let
...
   stableShell = pkgs.mkShellNoCC {
     packages = devDependencies ++ [ bky-as-stable ];
+    shellHook = ''
+      export AS_VERSION=${version}
+      render-md() {
+        for file in $(find . -type f -name '*.md'); do
+          mo "$file" > "$file.tmp" && mv "$file.tmp" "$file"
+        done
+      }
+    '';
```
matteobjornsson, line 32: I think this PR is great. My only concern is that this function feels a little "hidden in the nix configuration". A lot of times we see things like this in the project makefile. We haven't really crossed this bridge yet, but I am hoping that dev commands find a central, self documenting location. Either in a makefile or somewhere else that we can point to. This is also bringing up the idea that there are certain things we can do in the stable shell that we don't want to happen in the unstable shell or non-nix shell. This makes me wonder if we might want a mechanism to add to scripts that detects the shell environment. An example of maybe what it would look like to put this in the makefile:
```Makefile
render-templates:
	@if [ -z "$$AS_VERSION" ]; then \
		echo "Error: AS_VERSION is not set"; \
		exit 1; \
	fi
	for file in $(find . -type f -name '*.md'); do
	  	mo "$file" > "$file.tmp" && mv "$file.tmp" "$file"
	done
```

matteobjornsson (approved):
Caught up over video. Happy to see the tooling added for convenience when releasing and to iterate on how we want to support development before making some "conventions" decisions down the road.

coin_prices_from_coingecko/README.md
```diff
@@ -19,7 +19,7 @@ In this example, you'll learn how to:
 ## Setup
 
 - Install the Blocky AS CLI by following the
-  [setup instructions](https://blocky-docs.redocly.app/attestation-service/v0.1.0-beta.6/setup)
+  [setup instructions](https://blocky-docs.redocly.app/attestation-service/{{AS_VERSION}}/setup)
```
mwittie, line 22: I just noticed that the link to the error handling example above is missing a slash. I was going to issue a PR, but going through the repo, I only see the problem in this file, so might be quicker for you to just fix it here.

dlm: fixed.

nix/mkDevShell.nix
dlm: discussed over video call.  Plan is to hold off until we have a little more clarity on how we want to more generally handle the different tools for the different environments.

coin_prices_from_coingecko/README.md
```diff
@@ -8,7 +8,7 @@ Before starting this example, make sure you are familiar with the
 [Hello World - Attesting a Function Call](../hello_world_attest_fn_call/README.md),
 [Passing Input Parameters and Secrets](../params_and_secrets/README.md),
 and the
-[Error Handling - Attested Function Calls](../error_handling_attest_fn_callREADME.md)
+[Error Handling - Attested Function Calls](../error_handling_attest_fn_call/README.md)
```
dlm, line 11: fix missing `/`

PR #36, opened 2025-04-05
[BKY-6544] Allow specifying commit of unstable version

matteobjornsson:
## Describe your changes
While looking at the basm SDK and how we can manage the environment there using nix, the ability to specify a particular commit came up (for example, as we progress the SDK we may want to pin the tests against a specific commit of delphi, rather than just an official release or latest)

This PR suggests these changes here as a look into increasing the flexibility of the env management in the examples repo. 

## Related Jira cards
loosely: 
https://blocky.atlassian.net/browse/BKY-6544

## Notes for reviewers
We will have to update a few process notes. For example, previously when creating a release branch we would update "unstable" to the specific tag, where now we update "latest" to the specific tag. Nothing too crazy though. 

- [ ] I have run `make pre-pr` and all checks have passed.
    - We don't have a make pre-pr in this repo
    - Pretty sure
- [x] This PR is small. 
    - Otherwise, justify here:

### DIFF
--- a/nix/fetch-bky-as.sh
+++ b/nix/fetch-bky-as.sh
@@ -2,9 +2,10 @@
 
 set -e
 
-bin=$1
-os=$2
-arch=$3
+commit=$1
+bin=$2
+os=$3
+arch=$4
 
 mkdir -p $bin
 
@@ -17,13 +18,15 @@ else
     echo "no current version"
 fi
 
-echo -n "Getting the latest commit..."
-latest_commit=$(gh api repos/blocky/delphi/commits --jq '.[0].sha')
-echo "'$latest_commit'"
+if [[ $commit == "latest" ]]; then
+  echo -n "Getting the latest commit..."
+  commit=$(gh api repos/blocky/delphi/commits --jq '.[0].sha')
+  echo "'$commit'"
+fi
 
-if [[ $current_version_commit != $latest_commit ]]; then
+if [[ $current_version_commit != "$commit" ]]; then
     echo "Versions differ ...updating"
-    aws s3 cp s3://blocky-internal-release/delphi/cli/$latest_commit/${os}_${arch} ${bin}/bky-as
+    aws s3 cp "s3://blocky-internal-release/delphi/cli/${commit}/${os}_${arch}" "${bin}/bky-as"
     chmod +x "$bin/bky-as"
 else
     echo "Version up to date"
--- a/nix/mkDevShell.nix
+++ b/nix/mkDevShell.nix
@@ -8,6 +8,8 @@ let
   goos = system.goos;
   goarch = system.goarch;
 
+  isCommit = x: builtins.match "^[0-9a-f]{40}$" x != null;
+
   bky-as-stable = pkgs.stdenv.mkDerivation {
     pname = "bky-as";
     version = version;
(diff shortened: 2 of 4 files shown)


nix/fetch-bky-as.sh
```diff
@@ -17,13 +18,15 @@ else
...
 fi
 
-echo -n "Getting the latest commit..."
-latest_commit=$(gh api repos/blocky/delphi/commits --jq '.[0].sha')
-echo "'$latest_commit'"
+if [[ $commit == "latest" ]]; then
+  echo -n "Getting the latest commit..."
+  commit=$(gh api repos/blocky/delphi/commits --jq '.[0].sha')
+  echo "'$commit'"
+fi
```
matteobjornsson, line 25: How do you feel about this approach to grabbing the latest sha? We pass in the commit as the string "latest" and then fetch that commit. Otherwise we assume the commit being passed in is a valid git commit.

nix/mkDevShell.nix
```diff
@@ -22,11 +24,14 @@ let
 
   stableShell = pkgs.mkShellNoCC {
     packages = devDependencies ++ [ bky-as-stable ];
+    shellHook = ''
+      echo "Stable bky-as version: ${version}"
+    '';
```
matteobjornsson, line 29: I added these echo statements because when you start the nix shell a few times nothing prints out and it can be easy to forget what version you're running.

nix/system.nix
```diff
@@ -1,8 +1,8 @@
 { pkgs }:
 let
-  system = builtins.split "-" pkgs.stdenv.hostPlatform.system;
+  system = pkgs.lib.strings.splitString "-" pkgs.stdenv.hostPlatform.system;
   arch = builtins.elemAt (system) 0;
-  os = builtins.elemAt (system) 2;
+  os = builtins.elemAt (system) 1;
```
matteobjornsson, line 5: pkgs.lib.strings.splitString seemed like the right function here

shell.nix
```diff
@@ -1,3 +1,12 @@
+{
+  # Specify version from the command line by using a valid semver tag to grab a stable version, e.g.
+  #   `nix-shell --argstr version "v0.1.0-beta.5"`
+  # or a full git commit sha to grab an unstable version, e.g.
+  #   `nix-shell --argstr version "<full git commit sha>"`
+  # or use the default value of "latest" to get the latest unstable version, e.g.
+  #   `nix-shell`
+  version ? "latest",
```
matteobjornsson, line 8: this is the line we update when making a release branch now

nix/mkDevShell.nix
```diff
@@ -46,9 +51,15 @@ let
       ];
     shellHook = ''
       bin=$(pwd)/tmp/bin
-      fetch-bky-as.sh $bin ${goos} ${goarch}
+      fetch-bky-as.sh ${version} $bin ${goos} ${goarch}
```
dlm, line 54: FWIW, I would probably order the arguments as:
```
fetch-bky-as.sh $bin ${version} ${goos} ${goarch}
```
so that the arguments that  define the thing to fetch are all together.

nix/fetch-bky-as.sh
```diff
@@ -2,9 +2,10 @@
 
 set -e
 
-bin=$1
-os=$2
-arch=$3
+commit=$1
```
dlm, line 5: See comment below about argument order.

dlm (approved):
This is great!!  :shipit:

shell.nix
```diff
@@ -14,7 +23,7 @@ mkDevShell {
   # development shell. On release branches, such as "release/v0.1.0-beta.4"
   # this value should be "v0.1.0-beta.4".  On main, it should be set to
   # "unstable"
-  version = "unstable";
+  version = version;
```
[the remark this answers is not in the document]
matteobjornsson: ah, good point, thank you for catching that. Updated.

nix/fetch-bky-as.sh
matteobjornsson: updated as you suggested

PR #38, opened 2025-04-07
[Chore] update readme in the twap example with the latest code

mwittie:
## Describe your changes

This PR updates the twap readme with the latest code, that wasn't included in the readme in a previous PR

- [x] This PR is small. 
    - Otherwise, justify here:

### DIFF
--- a/time_weighted_average_price/attest_fn_call/README.md
+++ b/time_weighted_average_price/attest_fn_call/README.md
@@ -194,7 +194,11 @@ func iteration(inputPtr uint64, secretPtr uint64) uint64 {
 		return WriteError(outErr)
 	}
 
-	priceSamples, err := extractPriceSamples(args.EAttest, args.TAttest, args.Whitelist)
+	priceSamples, err := extractPriceSamples(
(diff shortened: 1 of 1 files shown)


matteobjornsson (approved):
Looks good. My understanding is this is a copy-paste of the code already existing on `main`. Feel free to auto-approve "keep the readme the same as the code" copy/paste PRs (IMO). I'm still a little out of the loop though which branch is the correct branch to point this PR at. Is this going to get cherry-picked back to `release/v0.1.0-beta.6`?

tahardi (approved):

tahardi:
@matteobjornsson I think this belongs in main since the `twap/attest_fn_call/main.go` function does in fact match this update to the docs. It can be cherry-picked into the `beta.6` release if needed.

If this doc update had been specific to `beta.6` then we would have merged directly to the release branch. For example, if a function definition was changed in `beta.6`, but not updated in the docs and was then changed on main right after release to something else. We would want to update the documentation in `beta.6` to match the function at that time but would not necessarily want it on main since it's no longer correct.

PR #41, opened 2025-04-13
[BKY-6576, BKY-6577] Update on chain to use ABI encoding and single transaction

dlm:
## Describe your changes

- update on-chain examples to use single transaction
- update on-chain examples to use ABI encoding of transitive attestations
- update all examples to use basm beta.8 
- fix typo in example Makefiles: 	`@echo "Running WASM **module..."` --> `@echo "Running WASM module..."`

## Related Jira cards

https://blocky.atlassian.net/browse/BKY-6576
https://blocky.atlassian.net/browse/BKY-6577

## Notes for reviewers

I have in the gas used in case you want to play around with it,
but I suspect that we will want to remove it.

Note that I have not made these changes in other contracts 
and have only run the `test-local` target.

- [x] This PR is small. 
    - Otherwise, justify here:


hello_world_on_chain/contracts/User.sol
```diff
@@ -19,18 +28,23 @@ contract User {
...
         claims = TAParserLib.verifyTransitivelyAttestedFnCall(
-                applicationPublicKeyAsAddress,
-                transitiveAttestation
-            );
+            applicationPublicKeyAsAddress,
+            transitiveAttestation
+        );
+
+        uint256 endGas = gasleft();
+        console.log("Gas used: %s", startGas-endGas);
```
dlm, line 39: Note that this produces the output:
```
Gas used: 14423                                                                                                                                                                                                                             ```

File: hello_world_on_chain/lib/TAParserLib.sol [hunk 2]
```diff
@@ -1,23 +1,21 @@
 //  SPDX-License-Identifier: MIT
 pragma solidity ^0.8.10;
 
-import {JsmnSolLib} from "./JsmnSolLib.sol";
```
dlm, line 4: I only made these changes to the version in hello world.  If we decide to go with this direction, we can add these changes to the other repos.

hello_world_on_chain/contracts/User.sol
mwittie: Sick!

mwittie: We should probably take this out to the output though.

mwittie (approved):
Looks good. Let me know when its in the final form and I'll do the Sepolia stuff and udpate the links to the chain explorer and the image of the logs.

I think I could also show the gas of the transaction on chain and mention how inexpensive it is based on the data from the block explorer.

dlm:
@mwittie sounds good.  WRT the gas, as mentioned in the PR description:

> I have in the gas used in case you want to play around with it,
> but I suspect that we will want to remove it.

I will take your comment to mean that I can now remove it.

PR #48, opened 2025-05-09
[BKY-6740] Initial tooling for testscript example tests

matteobjornsson:
## Describe your changes
This PR 
- adds a `/test` package for go-style test using `testscript`
- adds `test/testscript.go` for testscript test setup
- adds `test/examples_test.go` for per-example tests that use the example directories largely as-is.
- adds `test/live/live_test.go` for overwriting the config and running tests against a live server, such as staging.
- introduces minimal modifications to the `coin_prices_from_coingecko` example to allow for secret injection.

You can still run `make run` as you would previously, but to run the tests you will have to provide some values via environment variables, either via your shell configuration/rc, or inline. For example: 

```shell
YOUR_COINGECKO_API_KEY=$COINGECKO_API_DEV_KEY make test
```
or, to run the live test against staging:
```shell
LIVE_TEST_PLATFORM=nitro \
LIVE_TEST_CODE=${MEASURE} \
LIVE_TEST_AUTH_TOKEN=${TYK_API_DEV_KEY} \
LIVE_TEST_HOST="https://api.bky.sh/staging/delphi" \
YOUR_COINGECKO_API_KEY=${COINGECKO_API_DEV_KEY} \
make test-live
```
or, for nix-shell in `--pure` mode:
```
nix-shell \
  --pure \
  --argstr bkyAsVersion "latest" \
  --run "\
    YOUR_COINGECKO_API_KEY=${COINGECKO_API_DEV_KEY} \
    make pre-pr"
```

## Related Jira cards
https://blocky.atlassian.net/browse/BKY-6740

## Notes for reviewers
This PR sets the scaffolding for automated testing for future examples as well. High-level feedback on workflow and tooling and setup is very welcome. I have added the discussion label as an indicator that it might be more productive to get in touch to review this PR together if you do have a lot of high level feedback or questions.

Env variable management seems like it is in a good enough place to allow for fairly easy testing in CI, though it is annoying for dev workflows. This seemed quite out of scope for this PR, so env var management is simply expected to exist and be managed by the developer. 

If you could run this on your machine to make sure everything is working correctly, that would be great. If you need a coingecko API key you can either make one [here](https://www.coingecko.com/) or I can share one with you via lastpass.
```shell
nix-shell \
  --pure \
  --argstr bkyAsVersion "latest" \
  --run "\
    YOUR_COINGECKO_API_KEY=${COINGECKO_API_DEV_KEY} \
    make pre-pr"   
```

    Yes! 
    ```shell
    nix-shell \
      --pure \
      --argstr bkyAsVersion "latest" \
      --run "\
        YOUR_COINGECKO_API_KEY=${COINGECKO_API_DEV_KEY} \
        make pre-pr"   
    ```
- [x] This PR is small.
    - Otherwise, justify here:
    - Otherwise, describe requirements here: Maybe. Feel free to reach out to chat, some discussion over vc may be easier.


test/coin_prices_from_coingecko_test.go
```diff
@@ -0,0 +1,27 @@
...
+	"testing"
+
+	"github.com/rogpeppe/go-internal/testscript"
+)
+
+func TestCoinPricesFromCoingecko(t *testing.T) {
+	exampleName := "coin_prices_from_coingecko"
+	exampleDir := filepath.Join("..", exampleName)
+
+	work := newTestSetup(t, exampleDir).
```
dlm, line 14: I like the use of the fluent interface.  It is funny though, I almost want the DSL to encapsulate testscript all together so that the DSL describes what I want to do (not how I want to setup testscript).  I would try to explore something like:

```
newTest(exampleDir, exampleName).
    MakeTarget("build").
    CopyFile("config.toml").
    RenderTemplateFromEnvWithCleanup("fn-call.json", "YOUR_COINGECKO_API_KEY").
    CopyFile("tmp/x.wasm").
    RunTestscript(t)
````

A few comments on the changes:
1. Run make changed to make target, which helps to make the variables clear from the just call chain
2. CopyFile has only one arg.  Since most examples just copy from source to dest without changing the name, we can use the `CopyFile` to do that and then the user doesn't have to figure out if it is `CopyFile(src, dst)` or `CopyFile(dst, src)`... if we wanted something that changed the name we could add another function
3. RenderFile updated to `RenderTemplateFromEnvWithCleanup(src, envKeys...)` is a longer name, but it gets rid of the question of what the boolean is (if the default is to always cleanup, maybe we can drop the cleanup).  The `FromEnv` gives some hint to what the keys mean.  Finally, we can use varadic args so that we don't have to add the []string{})
4. Finally, we roll all of the testscript stuff into the DSL and then the user is just declaring a collection of actions.

Note that I am saying that you need to do any of this, just giving my two cents in where I would try to take this direction but YMMV.

File: test/testSetup.go [hunk 2]
```diff
@@ -0,0 +1,120 @@
...
+			}
+			envMap[key] = val
+		}
+
+		tmpl, err := template.New(filepath.Base(src)).Option("missingkey=error").ParseFiles(src)
+		if err != nil {
+			return fmt.Errorf("failed to parse template %s: %v", src, err)
+		}
+
+		// Create a buffer to store the rendered content
```
dlm, line 93: Do you need to write to the buffer and then from the buffer write to the file?  I believe that `Execute` just needs an io.Writer and so I think you should be able to write to the file directly.

test/testSetup.go
```diff
@@ -0,0 +1,120 @@
...
+	})
+	return ts
+}
+
+func (ts *testSetup) RenderFile(relPath string, envKeys []string, cleanup bool) *testSetup {
+	ts.setupFuncs = append(ts.setupFuncs, func(t *testing.T, workDir string) error {
+		src := filepath.Join(ts.srcDir, relPath)
+		dst := filepath.Join(workDir, filepath.Base(relPath))
+
+		// Prepare environment variables as a map
```
dlm, line 78: FWIW, some of these comments seem a little superfluous to me.  I don't think all of the comments add nothing, however, many are not that useful.  For example on line 102 `Write the rendered content to the destination file`, doesn't add a whole lot of information to the line `os.WriteFile(dst, buf.Bytes(),...)`  In fact, it does raise the question of "why not call `buf` something like `renderedTemplate` and then it would read `os.WriteFile(dst, renderedTemplate.Bytes()...)`.

But... I would put that as pretty nit picky... so up to you.

test/coin_prices_from_coingecko_test.go
matteobjornsson: I liked your suggestions, I rolled them into the PR with a few adjustments. Thank you for the feedback!

test/testSetup.go
matteobjornsson: nope, these were some leftovers from some "Vibe coding" sketches. They've been cleaned up. Thanks for the callout though about the bytes buffer, I forgot that file types in go are io.ReadWriters. Updated to skip the in-memory buffer

shell.nix
```diff
@@ -37,7 +37,7 @@ mkDevShell {
     pkgs.jq # for processing data in examples
     pkgs.nixfmt-rfc-style # for formatting nix files
     pkgs.nodejs_18 # for on chain examples
-    pkgs.tinygo # for building wasm
+    pkgs.docker # for building wasm
```
matteobjornsson, line 40: I'm no longer seeing tinygo as a dependency, just docker to pull the tinygo container and use it to run tinygo.

coin_prices_from_coingecko/README.md
```diff
@@ -86,7 +86,7 @@ using the `bky-as` CLI by passing in the
     "coin_id": "bitcoin"
   },
   "secret": {
-    "api_key": "CoinGecko API key"
+    "api_key": "{{.YOUR_COINGECKO_API_KEY}}"
```
matteobjornsson, line 89: README updated to reflect the small changes I made in the coingecko example.

coin_prices_from_coingecko/output.go
```diff
@@ -9,9 +9,9 @@ import (
 )
 
 type Result struct {
-	Success bool
-	Error   string
-	Value   any
+	Success bool   `json:"success"`
+	Error   string `json:"error,omitempty"`
+	Value   any    `json:"value,omitempty"`
```
matteobjornsson, line 14: added some json tags to trend away from go-case json fields.

test/examples_test.go
```diff
@@ -0,0 +1,35 @@
...
+	requiredEnvVars := []string{
+		"YOUR_COINGECKO_API_KEY",
+	}
+
+	NewProjectTest(t, projectDir).
+		ExecuteMakeTarget("build").
+		CopyFile("tmp/x.wasm").
+		CopyFile("config.toml").
+		RenderTemplateFileFromEnvWithCleanup("fn-call.json", requiredEnvVars).
+		RunScript(filepath.Join(scriptDir, projectName+".txtar"))
```
matteobjornsson, line 23: This setup expects a single script. This uses the `testscript.Params.Files` field under the hood. We could easily update this to run multiple scripts in the same testscript WorkDir if we wanted to. I did not see a big reason for this at the moment, so I left it as a "one shot"

test/examples_test.go
```diff
@@ -0,0 +1,35 @@
...
+const scriptDir = "scripts"
+
+func TestCoinPricesFromCoingecko(t *testing.T) {
+	projectName := "coin_prices_from_coingecko"
+	projectDir := filepath.Join(examplesDir, projectName)
+	requiredEnvVars := []string{
+		"YOUR_COINGECKO_API_KEY",
+	}
+
+	NewProjectTest(t, projectDir).
```
matteobjornsson, line 18: All testscript tests occur in a temporary working directory. These commands specify test and WorkDir setup actions that need to take place. For example, all "copy" and "render" commands move the specified file to the workdir.

test/live/live_test.go
```diff
@@ -0,0 +1,41 @@
...
+	"attestation-service-examples.test"
+)
+
+const liveTestConfigTemplate = `
+acceptable_measurements = [
+{ platform = "{{.LIVE_TEST_PLATFORM }}", code = "{{.LIVE_TEST_CODE}}" },
+]
+auth_token = "{{.LIVE_TEST_AUTH_TOKEN}}"
+host = "{{.LIVE_TEST_HOST}}"
+`
```
matteobjornsson, line 16: I like the fact that the non-live tests use each example project as-is. To perform live tests, though, we need to "override" the config. So, in the `live_test` package I supply a config template that we can populate using environment variables. It is then rendered and saved as a file in the testscript working directory.

test/scripts/hello_world_attest_fn_call.txtar
```diff
@@ -0,0 +1,10 @@
+# [execute] call the bky-as CLI to attest a function call
+stdin fn-call.json
+exec bky-as attest-fn-call
+cp stdout result.json
+
+# [check] assert stderr and stdout expected values
+exec jq -r '.transitive_attested_function_call.claims.output | @base64d' result.json
+stdout 'Hello, World!'
+exec jq -r '.transitive_attested_function_call.logs | @base64d' result.json
+stdout 'Writing "Hello, World!" to host'
```
matteobjornsson, line 10: here we see a little more of the testscript DSL. `stdout 'pattern'` just checks that the pattern is found in stdout, allowing us to assert some expected string/output

matteobjornsson: this testscript is slightly different from coingecko because the coingecko example uses a result type which needs to be parsed first. This example writes a string directly to the function output.

test/testscript.go
```diff
@@ -0,0 +1,155 @@
...
+
+	"github.com/rogpeppe/go-internal/testscript"
+)
+
+type ProjectTest struct {
+	t          *testing.T
+	params     testscript.Params
+	setupFuncs []func(*testscript.Env) error
+	projectDir string
+}
```
matteobjornsson, line 19: I called this struct a "ProjectTest" as the idea is it helps the user test a given project that have files/templates/artifacts/make targets we care about. 

Feel free to suggest naming updates.

test/testscript.go
```diff
@@ -0,0 +1,155 @@
...
+
+func NewProjectTest(t *testing.T, projectDir string) *ProjectTest {
+	params := testscript.Params{
+		Files:               []string{},
+		Setup:               nil,
+		RequireExplicitExec: true,
+		RequireUniqueNames:  true,
+	}
```
matteobjornsson, line 27: some default testscript parameters.

test/testscript.go
```diff
@@ -0,0 +1,155 @@
...
+			msg := "failed to create destination directory %s: %w"
+			return fmt.Errorf(msg, dstDir, err)
+		}
+
+		srcContent, err := os.ReadFile(src)
+		if err != nil {
+			return fmt.Errorf("failed to read source file %s: %w", src, err)
+		}
+
+		if err := os.WriteFile(dst, srcContent, 0644); err != nil {
```
matteobjornsson, line 50: I picked a permission mode of 644 here, for the files getting sent to the testscript workdir. We could also surface the permissions higher up, or pick a more restrictive set. LMK if you have any thoughts.

test/testscript.go
```diff
@@ -0,0 +1,155 @@
...
+		t:          t,
+		params:     params,
+		projectDir: projectDir,
+	}
+}
+
+func (e *ProjectTest) CopyFile(srcRelPath string) *ProjectTest {
+	setupFunc := func(env *testscript.Env) error {
+		src := filepath.Join(e.projectDir, srcRelPath)
+		dst := filepath.Join(env.WorkDir, srcRelPath)
```
matteobjornsson, line 38: for now, we recreate the given path from the project dir to the testscript workdir

test/testscript.go
```diff
@@ -0,0 +1,155 @@
...
+	templateString string,
+	dstRelPath string,
+	envKeys []string,
+) *ProjectTest {
+	tmpl := template.Must(
+		template.New(filepath.Base(dstRelPath)).
+			Parse(templateString),
+	)
+	return e.RenderTemplateFromEnvWithCleanup(tmpl, dstRelPath, envKeys)
+}
```
matteobjornsson, line 80: these Render wrappers allow for the generation of go templates from either a file (such as `fn-call.json` in the coingecko example) or from a string (such as the config template in `test/live/live_test.go`).

test/testscript.go
```diff
@@ -0,0 +1,155 @@
...
+		}
+		defer dstFile.Close()
+
+		tmpl.Option("missingkey=error")
+		if err := tmpl.Execute(dstFile, envMap); err != nil {
+			return fmt.Errorf("failed to execute template %s: %v", tmpl.Name(), err)
+		}
+
+		// rendered files may contain secrets, register a cleanup func for removal
+		env.Defer(func() {
```
matteobjornsson, line 117: note this uses `testscript.Env.Defer()`. When using `t.Cleanup()` the workdir will have already been deleted. There is still a risk of secrets being rendered into a temp directory if go test crashes. I think what we have handles the secret files sufficiently, but let me know what you think.

test/testscript.go
```diff
@@ -0,0 +1,155 @@
...
+	return e
+}
+
+func (e *ProjectTest) ExecuteMakeTarget(target string) *ProjectTest {
+	setupFunc := func(env *testscript.Env) error {
+		cmd := exec.Command("make", target)
+		cmd.Dir = e.projectDir
+		cmd.Stdout = os.Stdout
+		cmd.Stderr = os.Stderr
+		e.t.Logf("Running command: '%s'", cmd.String())
```
matteobjornsson, line 134: I added a log here so the user understands that a command is being run as part of the test. This is used to run `make build` in the example projects.

test/testSetup.go
matteobjornsson: Thanks for the callout though about the bytes buffer, I forgot that file types in go are io.ReadWriters. Updated to skip the in-memory buffer

coin_prices_from_coingecko/README.md
```diff
@@ -250,20 +250,19 @@ You'll see output similar to the following:
...
-  "Value": {
+  "success": true,
+  "value": {
     "market": "Binance",
     "coin_id": "BTC",
     "currency": "USD",
-    "price": 83799,
-    "timestamp": "2025-03-04T03:20:35Z"
+    "price": 104273,
+    "timestamp": "2025-05-13T21:25:01Z"
```
matteobjornsson, line 259: I re-ran the example for the new json field printout

README.md
```diff
@@ -59,3 +59,52 @@ following order:
...
+For example:
+
+```bash
+LIVE_TEST_PLATFORM=nitro \
+LIVE_TEST_CODE=${MEASURE} \
+LIVE_TEST_AUTH_TOKEN=${TYK_API_DEV_KEY} \
+LIVE_TEST_HOST="https://api.bky.sh/staging/delphi" \
+YOUR_COINGECKO_API_KEY=${COINGECKO_API_DEV_KEY} \
+make test-live
+```
```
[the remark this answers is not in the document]
tahardi: OHHHHHH I GET IT NOW

You didn't wrap that `make test-live` example command in a nix shell so it was using my local (old) version of `bky-as` which did not work with the version on staging. This scenario was the whole reason y'all added nix.. nice. `make test live` works fine when run in nix shell

README.md
```diff
@@ -59,3 +59,52 @@ following order:
...
+
+For example, using nix-shell to run tests in pure mode:
+
+```bash
+nix-shell \
+  --pure \
+  --argstr bkyAsVersion "latest" \
+  --run "\
+    YOUR_COINGECKO_API_KEY=${COINGECKO_API_DEV_KEY} \
+    make test"
```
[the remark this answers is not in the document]
tahardi: nvm this is from nix-shell running my local `bky-as inspect` which throws an error bc I don't have a config file in the directory. Thankfully it still works

test/testscript.go
tahardi: Probably fine since the owner has rw and it's all temporary anyway

test/testscript.go
tahardi: The one downside to `testscript` is that the output of `go test` is pretty cluttered. The statement you added here does not get printed because our Makefile omits the `-v` flag for `go test`. I added it back in to check and while I can see your log statement now there is also a bunch of output from `testscript`.

tahardi (approved):
This is awesome! Looks pretty intuitive (now that you've set it up at least). Everything worked on my system.

FYI Dave recommended the [direnv tool](https://direnv.net/) for managing environment variables. It's pretty useful in this scenario where we have per repo, or even per directory, env vars

Makefile
```diff
@@ -76,3 +76,14 @@ run-all: \
 	run-time \
 	run-params-and-secrets
 	@echo "All tests passed!"
+
+.PHONY: test
+test:
+	@go test -C test . -count=1
```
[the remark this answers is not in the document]
matteobjornsson: @tahardi this was actually harder to find in the docs than I expected. `-C` chdir, change directory
go 1.20 Changelog: https://tip.golang.org/doc/go1.20#go-command
> The go subcommands now accept -C <dir> to change directory to <dir> before performing the command, which may be useful for scripts that need to execute commands in multiple different modules.

README.md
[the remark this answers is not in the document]
matteobjornsson: yep, yep yep yep!! I went through the exact same process as you. For this reason I almost exclusively run make commands, pre-pr, etc, in nix now in Basm and examples repo since they depend on a specific version of `bky-as`

README.md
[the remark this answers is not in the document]
matteobjornsson: @tahardi I'm a little fuzzy on whether this is still a bug, for example in the `nix/fetch-bky-as.sh` script. I'm assuming this works for most people most of the time because they have a config somewhere that is preventing the cli tool from complaining. It would be nice if we could get the binary build details without needing a config

coin_prices_from_coingecko/README.md
mwittie: Thanks for making sure to update the readme. Could you also update the paragraph below to explain what the user needs to update (everything between the quotes) and why we are using this syntax in the config file. The explanation doesn't need to be super detailed.

matteobjornsson: yep, can do
- [ ] add explanation of template formatting

coin_prices_from_coingecko/output.go
mwittie: I think the omitempty could get confusing to users, when they don't see the field in the output. I think this works in our case because you're only testing for success in coin_prices_from_coingecko.txtar. However, omitting these values would change the output on README.md line 44, no?

test/testscript.go
matteobjornsson: yeah, the testscript package hooks into the go test framework so it is verbose if -v is used in `go test`, and testscript test output is quite verbose. So far I haven't figured out how to make the testscript output only "slightly verbose". I figure at the very least go switches to verbose output when things fail, so this log will at least be in there if the make command happens to be part of the problem. Open to suggestions/improvements for dev usability w.r.t. test output

mwittie (commented):
I went through the process and tested the pre-pr path. Very nice! 
I left a few small comments.

coin_prices_from_coingecko/output.go
matteobjornsson: hm, good point. I like the ergonomics of the omitempty behavior here as a dev, but it has changed a user facing thing and was not a necessary change. I can change it back if you think it is confusing behavior to users.

matteobjornsson:
@tahardi I figure env management is a big enough project on it's own I decided in convo with dave to punt on that, have it be its own PR. I will put an eng agenda item to prioritize/write a jira card. It might take some scoping work

dlm (approved):
This came out great!!

test/testscript.go
matteobjornsson: keeping as is for now!

README.md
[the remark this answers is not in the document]
matteobjornsson: resolving this comment as I moved this convo to slack and resolution of this convo is probably out of scope of this PR either way.

coin_prices_from_coingecko/README.md
matteobjornsson: done

coin_prices_from_coingecko/output.go
matteobjornsson: I removed the addition of omitempty

PR #50, opened 2025-05-16
[BKY-6811] Add testscript tests for go-only examples

tahardi:
## Describe your changes
This PR adds testscript tests for all of the go-only attestation service examples. That said, I left out the go portion of the TWAP example because it was more complicated than the rest and plan to address it along with the on-chain portion in a separate PR.

## Related Jira cards
[add tests for go-only examples](https://blocky.atlassian.net/browse/BKY-6811?atlOrigin=eyJpIjoiYThhNTk5Y2QwNWIzNDEwMGFhNTY0MTkzODFmYTVlY2YiLCJwIjoiaiJ9)

- [ ] I am merging into the intended base branch.
- [x] This PR is small.
    - Otherwise, justify here:


error_handling_attest_fn_call/fn-call-error.json
tahardi, line 1: All of the other examples name their request files `fn-call.json` so I updated to follow that format

esports_data_from_pandascore/fn-call.json
```diff
@@ -2,10 +2,9 @@
   "code_file": "tmp/x.wasm",
   "function": "scoreFunc",
   "input": {
-    "matches_api_endpoint": "PandaScore API matches endpoint",
```
tahardi, line 5: Didn't see the point in making the match API endpoint configurable given that it never changes. Also removing this as a config option made testing easier

test/scripts/error_handling_attest_fn_call.txtar
tahardi, line 1: @matteobjornsson what goland plugin were you using for these files? I noticed in your demo that it seems to syntax highlight yours but I didn't find any in the plugin manager

test/scripts/error_handling_attest_fn_call.txtar
```diff
@@ -0,0 +1,25 @@
+# [execute] function call that is successful
+stdin fn-call.json
+exec bky-as attest-fn-call
+cp stdout result.json
+
+# [check] assert stdout contains success output
+exec jq -c -r '.transitive_attested_function_call.claims.output | @base64d | fromjson' result.json
+stdout '{"Success":true,"Error":"","Value":{"number":42}}'
```
tahardi, line 8: Had to flatten the json because I couldn't figure out multi-line `stdout` testscript commands

test/scripts/error_handling_attest_fn_call.txtar
```diff
@@ -0,0 +1,25 @@
...
+exec bky-as attest-fn-call
+cp stdout result-error.json
+
+# [check] assert stdout contains error output
+exec jq -c -r '.transitive_attested_function_call.claims.output | @base64d | fromjson' result-error.json
+stdout '{"Success":false,"Error":"expected error","Value":null}'
+
+# [execute] function call that panics
+stdin fn-call-panic.json
+! exec bky-as attest-fn-call
```
tahardi, line 21: This is how you say you expect the command to fail

test/scripts/sports_data_from_sportradar.txtar
```diff
@@ -0,0 +1,8 @@
+# [execute] attest fn call to SportRadar API
+stdin fn-call.json
+exec bky-as attest-fn-call
+cp stdout result.json
+
+# [check] assert stdout contains correct game information
+exec jq -c -r '.transitive_attested_function_call.claims.output | @base64d | fromjson' result.json
+stdout '{"Success":true,"Error":"","Value":{"points_per_minute":'
```
tahardi, line 8: Not sure why, but whenever I provided the full string it would fail to find a match. I compared the output string to the one I provided in different editors and it said they were the same, but test script didn't seem to think so. Thus, I'm just searching for the first part of the sports radar return struct

test/scripts/error_handling_attest_fn_call.txtar
matteobjornsson: kind of random, but the atlas plugin
https://plugins.jetbrains.com/plugin/22307-atlas

test/scripts/error_handling_attest_fn_call.txtar
matteobjornsson: @tahardi for larger strings/files where you are matching the whole string rather than specifying a regex, you can include them as a file in the txtar archive and reference them, e.g. 
```txtar
-- expected-output.json --
{
  "Success": true,
  "Error": "",
  "Value": {
    "number": 42
  }
}
```
and I think you would use the `cmp` command to compare stdout to the file, e.g.
```txtar
cmp stdout expected-output.json
```

esports_data_from_pandascore/fn-call.json
matteobjornsson: are we allowed to include the pandascore API endpoint in these examples? I know that was a concern with coingecko, I wasn't sure if this "not specifying the API endpoint" was on purpose or not.

tahardi: It's a publicly documented API so I don't see why we couldn't. Also we do include the CoinGecko API directly in the code as opposed to reading in via a config

test/scripts/error_handling_attest_fn_call.txtar
tahardi: Interesting... It recommended that to me but the description had nothing to do with txtar so I ignored it

test/scripts/error_handling_attest_fn_call.txtar
tahardi: Thanks! I updated the examples with long output to use an `expected-output.json` file. Do you think I should do the same for every example regardless of output length?

esports_data_from_pandascore/fn-call.json
dlm: Kind of a strange, but even though that part of the API is publicly available, it is a violation of Pandascore's ToS for us to re-post parts of their API.  Kind of unfortunate as it creates a friction for their users, but it is nice that we can demo some of our system's features to address that friction. 

To address, please revert the changes to how the pandascore API is managed so that we go back to passing in the as a parameter.

matteobjornsson (changes requested):
From video chat discussion
- revert garden tending (or move to separate PR)
- use golden files for comparing expected deterministic string output (as part of txtar, not added to example dir)
- revert pandascore api string thing

esports_data_from_pandascore/fn-call.json
tahardi: Bummer. Removed explicit uses of the actual endpoint and replaced with templated env var instead

esports_data_from_rimble/Makefile
```diff
@@ -38,16 +38,14 @@ build: tmp/x.wasm
 .PHONY: match-winner
 match-winner: check build
 	@echo "Running WASM module..."
-	@sed -e 's/Rimble API key/$(RIMBLE_API_KEY)/' ./match-winner.json.template > ./tmp/match-winner.json
-	@cat ./tmp/match-winner.json | bky-as attest-fn-call >tmp/out.json
```
tahardi, line 42: The one garden tend-ish thing I left in was removing this templating bit from rimble example. Instead just let user modify the templated file by hand like the other examples

matteobjornsson: Is this change required for the PR? I am worried that this is reaching beyond the goal of this PR and possibly introducing breaking changes to workflows/processes or documentation (for example, I think Mike uses the makefile run targets). 

IMO of we are going to make changes that are for example tidying, unifying patterns between examples, or changing the behavior of the examples/example makefiles that are not required for test automation, we should push those to another PR or eng agenda to deal with those changes separately.

README.md
```diff
@@ -106,5 +106,9 @@ nix-shell \
   --argstr bkyAsVersion "latest" \
   --run "\
     YOUR_COINGECKO_API_KEY=${COINGECKO_API_DEV_KEY} \
+    YOUR_PANDASCORE_API_KEY=${PANDASCORE_API_DEV_KEY} \
```
matteobjornsson, line 109: does YOUR_PANDASCORE_API_ENDPOINT also need to be added to this list of env vars? Alternatively, if this example is more of a burden than a help, we can delete the example or provide a different one.

test/scripts/params_and_secrets.txtar
```diff
@@ -0,0 +1,17 @@
...
+stdout '{"Success":true,"Error":"","Value":{"decrypted_data":"your decrypted information"}}'
+
+# [execute] attest fn call that errors
+stdin fn-call-error.json
+exec bky-as attest-fn-call
+cp stdout result-error.json
+
+# [check] assert stdout does not contain decrypted information
+exec jq -c -r '.transitive_attested_function_call.claims.output | @base64d | fromjson' result-error.json
+stdout '{"Success":false,"Error":"decrypting data: incorrect password","Value":null}'
```
matteobjornsson, line 17: One thing that I think is ok for now but could become a thing in the future: My understanding is that golang serializes json in the same order as the struct, but this is an implementation detail and should not necessarily be relied on. For example, this discussion [here](https://github.com/golang/go/issues/6244#issuecomment-66084578). In using string matching rather than inspection with jq, we rely on the order. 

Again, I don't think this is something that needs to change, these tests are pretty straightforward, but it might be something to keep in mind and might be something we end up changing later.

matteobjornsson: Same thing goes for the output json files. Seems worth doing the way you've suggested for now though, much easier in getting this up and running and may never end up being something we need to be concerned with.

test/scripts/sports_data_from_sportradar.txtar
matteobjornsson, line 8:
```suggestion
exec jq -r '.transitive_attested_function_call.claims.output | @base64d | fromjson' result.json
cmp stdout expected-output.json

-- expected-output.json --
{
  "Success": true,
  "Error": "",
  "Value": {
    "points_per_minute": [
      {
        "player": "Jayson Tatum",
        "ppm": 0.689655172413793
      },
      {
        "player": "Luka Doncic",
        "ppm": 0.6454091432961967
      }
    ],
    "winner": "Jayson Tatum"
  }
}
```

matteobjornsson: not sure what you're running into. I wonder if there's something going on with newlines or the way that jq is producing the string with the `-c` flag. I suggested an edit below that uses the same pattern for comparing out put files instead.

test/scripts/time.txtar
```diff
@@ -0,0 +1,9 @@
+# [execute] attest fn call
+stdin fn-call.json
+exec bky-as attest-fn-call
+cp stdout result.json
+
+# [check] assert stdout contains correct output but
+# do not check the actual value as it is changes
+exec jq -c -r '.transitive_attested_function_call.claims.output | @base64d | fromjson' result.json
+stdout '{"Success":true,"Error":"","Value":'
```
matteobjornsson, line 9: If you wanted to match the full string you can use a pattern here, e.g. 
```
stdout '{"Success":true,"Error":"","Value":"[0-9]{4}-[0-9]{2}-[0-9]{2}T[0-9]{2}:[0-9]{2}:[0-9]{2}\.[0-9]{1,6}Z"}' # regex to match YYYY-MM-DDTHH:MM:SS.ssssssZ
```

esports_data_from_pandascore/fn-call.json
```diff
@@ -2,10 +2,10 @@
...
+    "matches_api_endpoint": "{{ .YOUR_PANDASCORE_API_ENDPOINT }}",
     "match_id": "1121861"
   },
   "secret": {
-    "api_key": "PandaScore API Key"
+    "api_key": "{{ .YOUR_PANDASCORE_API_KEY }}"
```
matteobjornsson, line 9: for the examples where you have modified files in the example (pandascore, dhl, sportsradar, rimble) can you check the readmes in those examples and update any relevant parts of the readmes to reflect your changes? There are likely parts of the readmes that instruct users on what strings to replace with their own values. Similar comment/request here as an example: https://github.com/blocky/attestation-service-examples/pull/48#discussion_r2089805701

matteobjornsson (approved):
Looks good. Some suggestions/comments, one suggestion to revert some garden tending, and one request to update example README.md files, but shouldn't hold up the PR I don't think. 

If you are using company wide secrets (for example, we have the sportsradar one shared, and I can make my own dhl and pandascore, but not rimble) can you add them to the LastPass? I ran what tests I could on my end.

matteobjornsson:
Oh, one other thing to consider: I think all these tests look like they could easily be converted over to live tests (replacing the config with one that points to staging) like in `test/live/live_test.go`. `live_test.go` represents our only true e2e test using the examples and basm and actual servers rather than the local-server versions of delphi. I think the goal is to continue adding tests over there, but I think it would be fine to do that in another PR or to revisit in eng agenda to discuss the future shape of e2e tests if it's starting to feel too redundant

esports_data_from_pandascore/fn-call.json
tahardi: done

test/scripts/sports_data_from_sportradar.txtar
tahardi: made this change myself while updating other files

PR #52, opened 2025-05-23
[BKY-6813] Hello World On Chain Testscript Test

tahardi:
## Describe your changes
This PR adds a testscript test for the "Local" test in the `hello_world_on_chain` example.

## Related Jira cards
[hello world on chain example testscript](https://blocky.atlassian.net/browse/BKY-6813?atlOrigin=eyJpIjoiZTBiMDVmMzJjOTUxNGVlMjhiNWU3Y2FlNDMwM2JiZDYiLCJwIjoiaiJ9)

## Notes for reviewers
This PR only adds a test for the "Local" on-chain test not the "Base Sepolia" live test version. That will be addressed in a future PR

- [x] This PR is small.
    - Otherwise, justify here:


test/testscript.go
```diff
@@ -32,27 +32,70 @@ func NewProjectTest(t *testing.T, projectDir string) *ProjectTest {
...
 		dst := filepath.Join(env.WorkDir, srcRelPath)
 		dstDir := filepath.Dir(dst)
-		if err := os.MkdirAll(dstDir, 0755); err != nil {
-			msg := "failed to create destination directory %s: %w"
-			return fmt.Errorf(msg, dstDir, err)
+
+		if err := makeDir(dstDir, 0755); err != nil {
+			return err
 		}
+		return copyFile(src, dst, 0644)
```
tahardi, line 63: What if we want to copy an executable file?

test/testscript.go
```diff
@@ -32,27 +32,70 @@ func NewProjectTest(t *testing.T, projectDir string) *ProjectTest {
 	}
 }
 
+func copyFile(srcPath string, dstPath string, mode os.FileMode) error {
```
tahardi, line 35: maybe a bit much to move these into separate funcs since they are so small but they were shared between CopyFile and the new CopyDir

test/testscript.go
```diff
@@ -32,27 +32,70 @@ func NewProjectTest(t *testing.T, projectDir string) *ProjectTest {
...
+			if err != nil {
+				return fmt.Errorf("creating relative path: %w", err)
+			}
+			dstPath := filepath.Join(dstDir, relPath)
+
+			if info.IsDir() {
+				return makeDir(dstPath, info.Mode())
+			} else {
+				return copyFile(srcPath, dstPath, info.Mode())
+			}
```
tahardi, line 89: Do you want me to hardcode the permissions like `CopyFile` or is it ok to use the permissions of the existing directories and files? What if we want to copy an executable file?

test/examples_test.go
```diff
@@ -23,6 +23,26 @@ func TestCoinPricesFromCoingecko(t *testing.T) {
...
+		CopyDir("inputs/").
+		CopyDir("lib/").
+		CopyDir("scripts/").
+		CopyDir("test/").
+		CopyFile(".env").
+		CopyFile("hardhat.config.ts").
+		CopyFile("package.json").
+		CopyFile("package-lock.json").
+		CopyFile("tsconfig.json").
+		SetEnv("TEST_CASE", "Local").
```
tahardi, line 41: This env var can be used to switch between the local ("Local") and live ("Base Sepolia") tests in the `hello_world_on_chain/test/user.ts` test file

test/testscript.go
```diff
@@ -141,6 +184,39 @@ func (e *ProjectTest) ExecuteMakeTarget(target string) *ProjectTest {
 	return e
 }
 
+func (e *ProjectTest) NPMInstallDeps() *ProjectTest {
```
tahardi, line 187: Used to install npm dependencies in the testscript workdir. It configures both `npm` and `npx` to write what would normally be global files to the testscript working directory instead so we don't muddy up the user's normal system

test/examples_test.go
matteobjornsson: Might be worth just hardcoding this for now and use a different script for "base sepolia" when later attempting to test live. There seems to be a trend towards keeping the testscripts as static as possible after setup (from what I've seen so far). Not necessarily something we need to do, but it might be worth slightly higher verbosity to have more clear-cut test conditions.

test/testscript.go
matteobjornsson: I had the same question when writing this but kicked it down the road. We could create another version of the copy fn that allows specification of the file permissions. Either that or change the testscript code to simply keep the existing file permissions which could be a simpler behavior.

test/testscript.go
matteobjornsson: Seems like keeping the original permissions might be the move. I was not sure if there is any reason to avoid it (tests will be run in other environments/runners, with secrets being written to files), but at the moment, looking at this again, it seems fine?

test/testscript.go
matteobjornsson: seems ok to me. Might be worth passing this by Dave to see if this is pushing the testing in any direction that he's not keen on. 

Either way, as far as getting from "not testable" to "at least we can run a test" seems like it does the job!

test/scripts/hello_world_on_chain.txtar
```diff
@@ -0,0 +1,5 @@
+# [execute]
+exec npx hardhat test --grep $TEST_CASE
```
matteobjornsson, line 2: does `pkgs.nodejs_18` cover all the node dependencies? I haven't run any of the node examples so I wasn't sure.

matteobjornsson (approved):
Looks good. Left some food for thought, and might be good to have dave take a look at the approach we are using for setting up the node projects, but I dig it!

test/testscript.go
```diff
@@ -141,6 +184,39 @@ func (e *ProjectTest) ExecuteMakeTarget(target string) *ProjectTest {
...
+	setupFunc := func(env *testscript.Env) error {
+		// Configure both npm and npx to only create and write files to the
+		// current working directory instead of the default $HOME directory
+		env.Setenv("NPM_CONFIG_CACHE", env.WorkDir+"/.npm")
+		env.Setenv("NPM_CONFIG_PREFIX", env.WorkDir+"/.npm-global")
+		env.Setenv("XDG_CACHE_HOME", env.WorkDir+"/.npx")
```
dlm, line 193: Also are you sure that these are correct?  This seems to be setting XDG confgs to npm/npx specifics folders.  For more on the XDG specification see https://specifications.freedesktop.org/basedir-spec/latest/

test/testscript.go
dlm: It does get the job done, but, I would strongly suggest against this approach as it can lead to some unexpected consequences.  In particular, this sets some `XDG_*` env vars, which a boarder scope than just nodejs.   That means that this function can produce some unexpected side effects that would be tricky to debug.  One example is that if there is another "step" in the ProjectTest that depends on the value of `XDG_CACHE_HOME`, it's behavior would change depending on if it was run before of after `NPMInstallDeps()`.

Generally, I try to decouple the env and the operations done in that env.  Ideally, I try to avoid env manipulation unless it is needed and if it is need, I try to leave it up the consumer of the function to make that decision.  So two options could be to move the env management to its own functions so that the setup looks more like:

```
NewProjectTest(t, projectDir).
                WithEphemerlXDGEnv().
                WithEphemeralNodeEnv().
		CopyDir("contracts/").
		...
```

Or better yet, let the decision be even higher, for example in the same place that setup the node environment or that sets up the environment for running this test. (That is, unless the point of the test is to do something novel with npm and xdg configs, which doesn't look to be the case.)  

On final comment, while the proposed solution does make sure that there are no files written to the user's directory, it is likely more aggressive than one would want.  Because it sets npm caches to the working dir, it doesn't allow npm to do any caching.  That means that every time a test that uses this step is run, it will need to re-download all npm dependencies, including when we are developing.  While that may be useful sometimes (ci system), it can make working on tools that use this step more challenging.

test/scripts/hello_world_on_chain.txtar
dlm: I would suggest against `node_18`.  It think it will work, but it's end of life was April 25, 2025 and so it is starting to be removed from many places.  I think I was using `node_22` when working on the sol based TAParser, but I don't know what version @mwittie was using.

dlm, line 2: I would suggest using `make test-local` to try to more closely follow what is in the writeup (and get a little coverage on the `Makefile`).

test/testscript.go
dlm: FWIW, I would suggest keeping the permissions as they were provided is a good default option.

test/testscript.go
```diff
@@ -141,6 +184,39 @@ func (e *ProjectTest) ExecuteMakeTarget(target string) *ProjectTest {
...
+		if err := cmd.Run(); err != nil {
+			return fmt.Errorf("failed to install npm deps: %v", err)
+		}
+		return nil
+	}
+	e.setupFuncs = append(e.setupFuncs, setupFunc)
+	return e
+}
+
+func (e *ProjectTest) SetEnv(key string, value string) *ProjectTest {
```
dlm, line 211: This setup function can probably be removed.  I think it was only used in examples_test.go:41.

dlm (changes requested):
Good first pass.  Let me know if you want to discuss some options, but we do need to do something wrt the env management.

test/scripts/hello_world_on_chain.txtar
tahardi: updated to use `make test-local`

test/testscript.go
tahardi: Updated to preserve file and directory permissions when copying

test/testscript.go
tahardi: updated to preserve file and directory permissions when copying

test/testscript.go
tahardi: removed

test/testscript.go
tahardi: Discussed over vc. Still setting `XDG` but it is more explicit now that it happens up in the test function.

test/scripts/hello_world_on_chain.txtar
tahardi: My system was 18, but I've updated to 22 and tested. Also bumped the nix shell version to 22 and that works as well

shell.nix
```diff
@@ -36,7 +36,7 @@ mkDevShell {
     pkgs.gotools # for tools like goimports
     pkgs.jq # for processing data in examples
     pkgs.nixfmt-rfc-style # for formatting nix files
-    pkgs.nodejs_18 # for on chain examples
+    pkgs.nodejs_22 # for on chain examples
```
tahardi, line 39: Dave pointed out node 18 is no longer supported. Bumped to 22 and tested

test/examples_test.go
```diff
@@ -23,6 +25,37 @@ func TestCoinPricesFromCoingecko(t *testing.T) {
...
 
+func TestHelloWorldOnChain(t *testing.T) {
+	projectName := "hello_world_on_chain"
+	projectDir := filepath.Join(examplesDir, projectName)
+	npmDir := func(env *testscript.Env) (string, error) {
+		return env.WorkDir+"/.npm", nil
+	}
+	npxDir := func(env *testscript.Env) (string, error) {
+		return env.WorkDir+"/.npx", nil
+	}
```
tahardi, line 36: In the first iteration, I was setting these environment variables inside of the `NPMInstall()` setup function. Dave advised against this, however, as it could lead to test misconfigurations if devs did not realize `NPMInstall()` was also changing the environment.

Setting these environment variables is now explicit, but I had to use this "lazy" evaluation of the value because I need the testscript working dir which is not actually available until runtime.

tahardi:
Discussed over video chat with Matteo and we decided that there are a lot of rough edges trying to run hardhat within the testscript framework. Instead, we will define a separate struct (and supporting functions) for running hardhat specific tests so that we can test our on-chain examples. Closing this PR and adding this changes in a new PR.

PR #54, opened 2025-06-04
[BKY-6813] On-Chain Tests

tahardi:
## Describe your changes
In PR #52 I tried adding an on-chain testscript test. Getting node and hardhat to work within the testscript framework, however, proved challenging. Ultimately, we decided that testscript was not suited for hardhat.

This PR renames the `ProjectTest` struct to the more appropriate `TestscriptTest` and introduces a `HardhatTest` struct for running `npm hardhat` tests.

- [x] This PR is small.
    - Otherwise, justify here:

### DIFF
--- a/shell.nix
+++ b/shell.nix
@@ -36,7 +36,7 @@ mkDevShell {
     pkgs.gotools # for tools like goimports
     pkgs.jq # for processing data in examples
     pkgs.nixfmt-rfc-style # for formatting nix files
-    pkgs.nodejs_18 # for on chain examples
+    pkgs.nodejs_22 # for on chain examples
     pkgs.docker # for building wasm
     pkgs.mo # for stamping version
   ];
--- a/test/examples_test.go
+++ b/test/examples_test.go
@@ -72,27 +72,27 @@ func TestESportsDataFromPandaScore(t *testing.T) {
 		Run(filepath.Join(scriptDir, projectName+".txtar"))
 }
 
-func TestESportsDataFromRimble(t *testing.T) {
-	projectName := "esports_data_from_rimble"
-	projectDir := filepath.Join(examplesDir, projectName)
-	requiredEnvVars := []string{
-		"YOUR_RIMBLE_API_KEY",
-	}
-
-	NewTestscriptTest(t, projectDir).
-		ExecuteMakeTarget("build").
-		CopyFile("tmp/x.wasm").
-		CopyFile("config.toml").
-		RenderTemplateFileFromEnvWithCleanup(
-			"match-winner.json.template",
-			requiredEnvVars,
-		).
-		RenderTemplateFileFromEnvWithCleanup(
-			"team-kill-diff.json.template",
-			requiredEnvVars,
-		).
-		Run(filepath.Join(scriptDir, projectName+".txtar"))
-}
+//func TestESportsDataFromRimble(t *testing.T) {
+//	projectName := "esports_data_from_rimble"
+//	projectDir := filepath.Join(examplesDir, projectName)
+//	requiredEnvVars := []string{
+//		"YOUR_RIMBLE_API_KEY",
+//	}
+//
+//	NewTestscriptTest(t, projectDir).
+//		ExecuteMakeTarget("build").
+//		CopyFile("tmp/x.wasm").
+//		CopyFile("config.toml").
+//		RenderTemplateFileFromEnvWithCleanup(
+//			"match-winner.json.template",
+//			requiredEnvVars,
+//		).
+//		RenderTemplateFileFromEnvWithCleanup(
+//			"team-kill-diff.json.template",
+//			requiredEnvVars,
+//		).
+//		Run(filepath.Join(scriptDir, projectName+".txtar"))
+//}
 
 func TestHelloWorldAttestFnCall(t *testing.T) {
 	projectName := "hello_world_attest_fn_call"
--- a/test/examples_test.go
+++ b/test/examples_test.go
@@ -20,7 +20,7 @@ func TestCoinPricesFromCoingecko(t *testing.T) {
 		CopyFile("tmp/x.wasm").
 		CopyFile("config.toml").
 		RenderTemplateFileFromEnvWithCleanup("fn-call.json", requiredEnvVars).
-		RunScript(filepath.Join(scriptDir, projectName+".txtar"))
+		Run(filepath.Join(scriptDir, projectName+".txtar"))
 }
 
 func TestErrorHandlingAttestFnCall(t *testing.T) {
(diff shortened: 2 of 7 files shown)


shell.nix
```diff
@@ -36,7 +36,7 @@ mkDevShell {
     pkgs.gotools # for tools like goimports
     pkgs.jq # for processing data in examples
     pkgs.nixfmt-rfc-style # for formatting nix files
-    pkgs.nodejs_18 # for on chain examples
+    pkgs.nodejs_22 # for on chain examples
```
tahardi, line 39: bump node version as 18 is no longer supported

test/examples_test.go
```diff
@@ -15,25 +15,45 @@ func TestCoinPricesFromCoingecko(t *testing.T) {
...
 		CopyFile("tmp/x.wasm").
 		CopyFile("config.toml").
 		CopyFile("successFunc.json").
 		CopyFile("errorFunc.json").
 		CopyFile("panicFunc.json").
-		RunScript(filepath.Join(scriptDir, projectName+".txtar"))
+		Run(filepath.Join(scriptDir, projectName+".txtar"))
+}
+
+func TestErrorHandlingOnChain(t *testing.T) {
```
tahardi, line 39: Went ahead and added on-chain tests for hello-world, error handling, and twap even though the card just called for hello-world. It wasn't hard and I thought it would be a better demonstration of how `HardhadTest` works to have multiple examples

test/examples_test.go
```diff
@@ -15,25 +15,45 @@ func TestCoinPricesFromCoingecko(t *testing.T) {
...
+	projectName := "error_handling_on_chain"
+	projectDir := filepath.Join(examplesDir, projectName)
+
+	expectedOutput1 := `Success: true
+	Error: 
+	Value: {"number":42}`
+	expectedOutput2 := `Success: false
+	Error: expected error
+	Value: null`
```
tahardi, line 48: error handling on chain runs two tests. These are the expected outputs for each

test/examples_test.go
```diff
@@ -15,25 +15,45 @@ func TestCoinPricesFromCoingecko(t *testing.T) {
...
+
+	NewHardhatTest(t, projectDir).
+		NPMInstall().
+		OutputContains(expectedOutput1).
+		OutputContains(expectedOutput2).
+		OutputContains("2 passing").
+		NoError().
+		Run("--grep", "Local")
```
tahardi, line 56: Specify the local test. Otherwise it will just run `npm hardhat test` which includes both local and live tests

test/examples_test.go
```diff
@@ -44,68 +64,87 @@ func TestESportsDataFromPandaScore(t *testing.T) {
...
+//		RenderTemplateFileFromEnvWithCleanup(
+//			"match-winner.json.template",
+//			requiredEnvVars,
+//		).
+//		RenderTemplateFileFromEnvWithCleanup(
+//			"team-kill-diff.json.template",
+//			requiredEnvVars,
+//		).
+//		Run(filepath.Join(scriptDir, projectName+".txtar"))
+//}
```
tahardi, line 95: Not sure why, but Rimble stopped working all of a sudden. Maybe they changed the API? Could you check on your end? I commented out and was going to make a card to investigate this outside of this PR

tahardi: Specifically, I was getting a 204 response which leads me to believe that my key still works but something about the particular api we are calling may have changed

test/examples_test.go
```diff
@@ -15,25 +15,45 @@ func TestCoinPricesFromCoingecko(t *testing.T) {
...
+
+	expectedOutput1 := `Success: true
+	Error: 
+	Value: {"number":42}`
+	expectedOutput2 := `Success: false
+	Error: expected error
+	Value: null`
+
+	NewHardhatTest(t, projectDir).
+		NPMInstall().
```
tahardi, line 51: Install deps first. Nice thing is that since we are using the examples directories deps will be cached for successive runs

test/hardhat.go
```diff
@@ -0,0 +1,84 @@
...
+	return h
+}
+
+func (h *HardhatTest) NoError() *HardhatTest {
+	checkFunc := func(_ string, err error) {
+		assert.NoError(h.t, err)
+	}
+	h.checkFuncs = append(h.checkFuncs, checkFunc)
+	return h
+}
```
tahardi, line 50: Maybe this is not needed and instead we always want to assert that `npm hardhat test` succeeded. On the other hand, this would allow us to test that misconfigurations fail. Additionally, we can use `OutputContains` to check the specific reason for failure

test/hardhat.go
```diff
@@ -0,0 +1,84 @@
...
+
+	"github.com/stretchr/testify/assert"
+)
+
+type HardhatTest struct {
+	t          *testing.T
+	checkFuncs []func(string, error)
+	setupFuncs []func() error
+	projectDir string
+}
```
tahardi, line 19: struct specifically for running `npm hardhat test`. We can add support setup functions on the struct for `hardhat compile`, `verify`, etc in the future if needed for the live test.

test/examples_test.go
```diff
@@ -44,68 +64,87 @@ func TestESportsDataFromPandaScore(t *testing.T) {
...
-			"match-winner.json.template",
-			requiredEnvVars,
-		).
-		RenderTemplateFileFromEnvWithCleanup(
-			"team-kill-diff.json.template",
-			requiredEnvVars,
-		).
-		RunScript(filepath.Join(scriptDir, projectName+".txtar"))
-}
+//func TestESportsDataFromRimble(t *testing.T) {
```
matteobjornsson, line 75: is this an accident that this is commented out?

test/examples_test.go
matteobjornsson: Great, agreed it's easier to set up test tooling/harness with a few examples

test/examples_test.go
matteobjornsson: Discussed over vc, it seems that we can leave the definition of "test passed" to the hardhat test, rather than checking for stdout string values

matteobjornsson (commented):
Discussed trimming down the hardhat tests to a very thin wrapper (leaving definition of test passing to the `hardhat test` command). Feel free to re-request review when changes are made!

test/examples_test.go
tahardi: Added TODO with card link

time_weighted_average_price/on_chain/package-lock.json
```diff
@@ -1858,7 +1858,8 @@
       "version": "1.0.9",
       "resolved": "https://registry.npmjs.org/abbrev/-/abbrev-1.0.9.tgz",
       "integrity": "sha512-LEyx4aLEC3x6T0UguF6YILf+ntvmOaWsVfENmIW0E9H09vKlLDGelMjjSm0jkDHALj8A8quZ/HapKNigzwge+Q==",
-      "license": "ISC"
+      "license": "ISC",
+      "peer": true
```
matteobjornsson, line 1862: What's this diff from? Is this from the new version of node?

matteobjornsson (approved):
Looks great. One question about the package lock diff

time_weighted_average_price/on_chain/package-lock.json
tahardi: ya bumping to node v22 caused this update

PR #55, opened 2025-06-06
[BKY-6883] Update Rimble match

matteobjornsson:
## Describe your changes
Update the rimble match in the example to a more recent match. The `204 No Content` HTTP response for the previous match really looks like it is no longer being served(?), possibly aged out and should have really been a `404 Not Found` response. 

Might be worth getting in touch with rimble to find out if there is a max age of match data they serve or something like that. 

## Related Jira cards
https://blocky.atlassian.net/browse/BKY-6883

## Notes for reviewers
I am not 100% sure that the cause is because it aged out, but other match IDs and dates worked for me. 

@mwittie tagging you because I do not think this changes anything in the docs but I wanted to make sure. 

- [x] This PR is small.
    - Otherwise, justify here:

### DIFF
--- a/esports_data_from_rimble/README.md
+++ b/esports_data_from_rimble/README.md
@@ -176,9 +176,9 @@ to get the output:
   "Success": true,
   "Error": "",
   "Value": {
-    "MatchID": "2379357",
-    "Date": "2025-02-18",
-    "Winner": "MOUZ"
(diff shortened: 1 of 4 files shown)


tahardi (approved):
Interesting find. The updated match info works for me as well!

PR #64, opened 2025-06-25
[BKY-6790] on schedule test

tahardi:
## Describe your changes
This PR adds an `on-schedule.yml` that runs a nightly test of our attestation service examples against a live `bky-as` server (as opposed to the local version used in `on-prs`)

## Related Jira cards
[on schedule test](https://blocky.atlassian.net/browse/BKY-6790?atlOrigin=eyJpIjoiYzMzODk4ZTBlYzg4NGQ5NDg3MGFlMTgyNTA5ZDE0YjUiLCJwIjoiaiJ9)

- [x] This PR is small.
    - Otherwise, justify here:

### DIFF
--- a/.github/workflows/on-schedule.yml
+++ b/.github/workflows/on-schedule.yml
@@ -1,0 +1,80 @@
+name: Scheduled Tests
+
+on:
+  push:
+  workflow_dispatch:
+  schedule:
+    - cron: '0 8 * * *'
+
+jobs:
+  test-with-live-bky-server:
+    name: Test Attestation Service Examples Against Live BKY-AS Server
+    runs-on: ubuntu-latest
+    permissions:
+      id-token: write # Required for OIDC
+      contents: read  # Required to clone repo
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Configure AWS credentials
+        uses: aws-actions/configure-aws-credentials@v4.1.0
+        with:
+          # Use the GH Actions role with limited permissions (i.e. read only access)
+          # defined here https://github.com/blocky/atlantis-aws/blob/main/aws_iam.tf
+          role-to-assume: arn:aws:iam::${{ vars.AWS_ACCOUNT_ID }}:role/BkyOidcGithubActionsBlocky
+          aws-region: us-west-2
+
+      - name: Setup Go
+        uses: actions/setup-go@v5
+        with:
+          go-version-file: './test/go.mod'
+
+      - name: Install Nix
+        uses: cachix/install-nix-action@v31
+        with:
+          nix_path: nixpkgs=channel:nixos-24.11
+
+      - name: Fetch Server Enclave Measurement
+        run: |
+          response=$(curl \
+            --fail \
+            -H "Authorization: ${{ secrets.TYK_GITHUB_WORKFLOW_API_TOKEN }}" \
+            https://api.bky.sh/staging/delphi/unsafe/inspect
+          )
+
+          server_code_measure=$(echo "$response" | jq -r '.enclave_measurement.code')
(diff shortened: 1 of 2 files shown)


.github/workflows/on-schedule.yml
```diff
@@ -0,0 +1,79 @@
...
+            --keep AWS_SESSION_TOKEN \
+            --keep AWS_REGION \
+            --keep LIVE_TEST_AUTH_TOKEN \
+            --keep LIVE_TEST_CODE \
+            --keep LIVE_TEST_HOST \
+            --keep LIVE_TEST_PLATFORM \
+            --keep GH_TOKEN \
+            --keep YOUR_COINGECKO_API_KEY \
+            --run "\
+            go test -C test ./live -count=1 -v -run=TestLiveCoinPricesFromCoingecko"
```
tahardi, line 79: Turns out we only have a single test in the "live" test suit right now. Should I make cards for adding others? Should I update this to just do `make test-live`?

.github/workflows/on-schedule.yml
tahardi, line 1: I tested this before putting up the pr by adding the `push` trigger so I could check that it ran correctly

.github/workflows/on-schedule.yml
```diff
@@ -0,0 +1,79 @@
...
+      id-token: write # Required for OIDC
+      contents: read  # Required to clone repo
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Configure AWS credentials
+        uses: aws-actions/configure-aws-credentials@v4.1.0
+        with:
+          # Use the GH Actions role with limited permissions (i.e. read only access)
+          # defined here https://github.com/blocky/atlantis-aws/blob/main/aws_iam.tf
```
matteobjornsson, line 22: this is outdated, I think it's supposed to point to this: https://github.com/blocky/atlantis-aws/blob/main/aws_iam_role.tf

This applies to the comment in the on-pr workflow as well

.github/workflows/on-schedule.yml
matteobjornsson: yes and yes. Though we could add an item to eng agenda to verify that the live tests are valuable to us before adding all of the examples to the live test. 

the organization of the examples tests in local vs live is a bit clunky/verbose/redundant right now. There is some room to maybe improve that, but we are still exploring what patterns make sense, so we don't want to try to pre-generalize too early and it could be ok to keep it verbose/repetative for now while we figure out what is working for us.

.github/workflows/on-schedule.yml
```diff
@@ -0,0 +1,79 @@
...
+          LIVE_TEST_PLATFORM: ${{ env.SERVER_PLATFORM }}
+          GH_TOKEN: ${{ secrets.GH_ACCESS_TOKEN }}
+          YOUR_COINGECKO_API_KEY: ${{ secrets.COINGECKO_API_KEY }}
+        run: |
+          nix-shell \
+            --pure \
+            --argstr bkyAsVersion "latest" \
+            --keep AWS_ACCESS_KEY_ID \
+            --keep AWS_SECRET_ACCESS_KEY \
+            --keep AWS_SESSION_TOKEN \
```
matteobjornsson, line 70: This is a little spooky to have in a public repo, if we were ever to accidentally log or leak these creds, but I think the OIDC short lived, scoped tokens make me feel wayyyyy better about it.

matteobjornsson (approved):
Looks good! Can you update the #monitoring channel to subscribe to the outcome of these on-schedule workflows in the same way we are subscribed to other on-schedule workflows? Happy to advise if you're not sure how.

PR #65, opened 2025-07-01
[BKY-6950] Update Rimble to use recent match id

tahardi:
## Describe your changes
This PR updates the Rimble example so that it first queries for a list of recent matches. It then uses the first in match in the list to perform the "match winner" and "team kill diff" functions. The reason for this change is that match data is only available for two weeks on our current price tier, so we can't hardcode our examples/tests to use a specific match id.

## Related Jira cards
[rimble recent match id](https://blocky.atlassian.net/browse/BKY-6950?atlOrigin=eyJpIjoiYzgxMjNjNzI3YTgxNDUyOWI5YTBmMjM4ZDdmODFkN2UiLCJwIjoiaiJ9)

- [x] This PR is small.
    - Otherwise, justify here:

### DIFF
--- a/esports_data_from_rimble/Makefile
+++ b/esports_data_from_rimble/Makefile
@@ -57,12 +57,6 @@ team-kill-diff: check build
 clean:
 	@rm -rf tmp
 
-.PHONY: update-rimble-test-data
-update-rimble-test-data:
-	@curl -s "https://rimbleanalytics.com/raw/csgo/match-status/?matchid=2382907&date=2025-06-03" \
-		-H "x-api-key: $(RIMBLE_API_KEY)" \
-		| jq . > rimble/testdata/match_data.json
-
 .PHONY: test-rimble
 test-rimble:
 	go test -v ./rimble/...
--- a/esports_data_from_rimble/main.go
+++ b/esports_data_from_rimble/main.go
@@ -9,18 +9,11 @@ import (
 	"github.com/blocky/basm-go-sdk/basm"
 )
 
-func getMatchDataFromRimble(
-	date string,
-	matchID string,
-	apiKey string,
-) (
-	rimble.MatchData,
-	error,
-) {
-	rimbleURL := "https://rimbleanalytics.com/raw/csgo/match-status/"
+func getRecentMatchesFromRimble(apiKey string) ([]rimble.MatchData, error) {
+	rimbleURL := "https://rimbleanalytics.com/raw/csgo/completed-matches/"
 	req := basm.HTTPRequestInput{
 		Method: "GET",
-		URL:    fmt.Sprintf("%s?matchid=%s&date=%s", rimbleURL, matchID, date),
+		URL:    rimbleURL,
 		Headers: map[string][]string{
 			"Accept":    {"application/json"},
 			"x-api-key": {apiKey},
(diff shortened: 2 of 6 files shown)


esports_data_from_rimble/Makefile
```diff
@@ -57,12 +57,6 @@ team-kill-diff: check build
 clean:
 	@rm -rf tmp
 
-.PHONY: update-rimble-test-data
```
tahardi, line 60: Since we don't pin our examples to a specific match id anymore I got rid of this convenience function. This really only needs to be updated when the match status response format changes anyway, which should be very infrequent

esports_data_from_rimble/main.go
```diff
@@ -74,42 +64,29 @@ type SecretArgs struct {
 	RimbleAPIKey string `json:"api_key"`
 }
 
-type MatchWinnerArgs struct {
```
tahardi, line 77: Neither `match-winner.json` or `team-kill-diff.json` fn call examples require inputs now that we are not passing in an expected date and matchID

esports_data_from_rimble/main.go
```diff
@@ -74,42 +64,29 @@ type SecretArgs struct {
...
+	case err != nil:
+		outErr := fmt.Errorf("getting recent matches: %w", err)
+		return WriteError(outErr)
+	case len(matches) == 0:
+		outErr := fmt.Errorf("no recent matches found")
 		return WriteError(outErr)
 	}
 
-	matchWinner, err := TeamWinner(match, input.Date)
+	matchWinner, err := TeamWinner(matches[0])
```
tahardi, line 87: both the "match winner" here and "team kill diff" logic further down behave the same except that we are using the first most recent match returned by Rimble as opposed to fetching one with a hardcoded match id

esports_data_from_rimble/rimble/rimble.go
```diff
@@ -50,6 +50,7 @@ type Metadata struct {
 
 type MatchData struct {
 	Metadata    Metadata `json:"metadata"`
+	Date        string   `json:"date"`
```
tahardi, line 53: This is actually included in the match data response structure and can be used instead of passing it what we expect the date to be

test/scripts/esports_data_from_rimble.txtar
```diff
@@ -5,7 +5,12 @@ cp stdout result-match-winner.json
 
 # [check] assert stdout contains correct match information
 exec jq -r '.transitive_attested_function_call.claims.output | @base64d | fromjson' result-match-winner.json
-cmp stdout expected-output-match-winner.json
+cp stdout output.json
+exec jq -e '.Success == true' output.json
+exec jq -e '.Error == ""' output.json
+exec jq -e '.Value.MatchID' output.json
+exec jq -e '.Value.Date' output.json
+exec jq -e '.Value.Winner' output.json
```
tahardi, line 13: A little hacky/brittle, but because the match data is subject to change I can only check for the expected field names not their actual values. I tried doing some fancy regex stuff I could not get commands with multi-line strings to work with testscript/txtar

esports_data_from_rimble/main.go
```diff
@@ -9,18 +9,11 @@ import (
...
-func getMatchDataFromRimble(
-	date string,
-	matchID string,
-	apiKey string,
-) (
-	rimble.MatchData,
-	error,
-) {
-	rimbleURL := "https://rimbleanalytics.com/raw/csgo/match-status/"
+func getRecentMatchesFromRimble(apiKey string) ([]rimble.MatchData, error) {
```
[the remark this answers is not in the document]
mwittie: See overall comment

mwittie (changes requested):
I think that this approach you take of querying for a recent match ID in the attested execution might a bit difficult to explain in the documentation write up of the example. The flow I had in mind is that the user finds a particular match they are interested in and then passes it to `match-winner.json` as a parameter to then get an attestation over that particular match. That's what I tried to express as the done condition "query for a recent game ID and template fn-call.json with a game ID" in https://blocky.atlassian.net/browse/BKY-6950. I thought that the game match lookup could be done using `curl` and the result templated into `match-winner.json.template` before even invoking `bky-as`. We could describe this pre-process in the documentation.

The way you coded this, we are creating an attestation over a random recent match, which shows users how to write a function that talks to Rimble, but the story around why it's useful to a potential user is a bit less clear.

Let me know if you want to connect on this, but I think we should keep the story clear.

tahardi:
@mwittie I'm closing this PR and opening a separate one that addresses your feedback

PR #66, opened 2025-07-09
[BKY-6950] Rimble fetch match id bky 6950

tahardi:
## Describe your changes
Rimble match data is only available for two weeks. Thus, we cannot hardcode the match id in our rimble example. This PR updates the rimble example so that the match id and match date are now templated (similar to the api key) we can dynamically specify the matchid at runtime.

## Related Jira cards
[recent rimble id](https://blocky.atlassian.net/browse/BKY-6950?atlOrigin=eyJpIjoiZjY3OWNmMzAyNWU3NGZkMGIwMzc2MzM5OTUzNWEyYzEiLCJwIjoiaiJ9)

- [x] This PR is small.
    - Otherwise, justify here:

### DIFF
--- a/test/scripts/esports_data_from_rimble.txtar
+++ b/test/scripts/esports_data_from_rimble.txtar
@@ -5,7 +5,12 @@ cp stdout result-match-winner.json
 
 # [check] assert stdout contains correct match information
 exec jq -r '.transitive_attested_function_call.claims.output | @base64d | fromjson' result-match-winner.json
-cmp stdout expected-output-match-winner.json
+cp stdout output.json
+exec jq -e '.Success == true' output.json
+exec jq -e '.Error == ""' output.json
+exec jq -e '.Value.MatchID' output.json
+exec jq -e '.Value.Date' output.json
+exec jq -e '.Value.Winner' output.json
 
 # [execute] attest fn call to Rimble API for team kill information
 stdin team-kill-diff.json.template
@@ -14,28 +19,11 @@ cp stdout result-team-kill-diff.json
 
 # [check] assert stdout contains correct team kill information
 exec jq -r '.transitive_attested_function_call.claims.output | @base64d | fromjson' result-team-kill-diff.json
-cmp stdout expected-output-team-kill-diff.json
-
--- expected-output-match-winner.json --
-{
-  "Success": true,
-  "Error": "",
-  "Value": {
-    "MatchID": "2382613",
-    "Date": "2025-06-20",
-    "Winner": "paiN"
-  }
-}
--- expected-output-team-kill-diff.json --
-{
-  "Success": true,
-  "Error": "",
-  "Value": {
-    "MatchID": "2382613",
-    "Date": "2025-06-20",
-    "MapName": "Inferno",
-    "Team1": "paiN",
-    "Team2": "FURIA",
-    "KillDiff": 12
-  }
-}
+cp stdout output.json
+exec jq -e '.Success == true' output.json
+exec jq -e '.Error == ""' output.json
+exec jq -e '.Value.MatchID' output.json
+exec jq -e '.Value.Date' output.json
+exec jq -e '.Value.Team1' output.json
+exec jq -e '.Value.Team2' output.json
+exec jq -e '.Value.KillDiff' output.json
--- a/test/examples_test.go
+++ b/test/examples_test.go
@@ -62,6 +62,8 @@ func TestESportsDataFromRimble(t *testing.T) {
 	projectName := "esports_data_from_rimble"
 	projectDir := filepath.Join(examplesDir, projectName)
 	requiredEnvVars := []string{
+		"YOUR_RIMBLE_MATCH_DATE",
+		"YOUR_RIMBLE_MATCH_ID",
 		"YOUR_RIMBLE_API_KEY",
 	}
 


esports_data_from_rimble/Makefile
```diff
@@ -1,4 +1,4 @@
-default: run
+default: match-winner
```
tahardi, line 1: `run` is not actually a target in this makefile (though it is in the other examples). I update to default to `match-winner` target

esports_data_from_rimble/Makefile
```diff
@@ -40,29 +40,36 @@ build: tmp/x.wasm
 .PHONY: match-winner
 match-winner: check build
 	@echo "Running WASM module..."
-	@sed -e 's/{{ .YOUR_RIMBLE_API_KEY }}/$(RIMBLE_API_KEY)/' ./match-winner.json.template > ./tmp/match-winner.json
+	@sed -e 's/{{ .YOUR_RIMBLE_MATCH_DATE }}/$(RIMBLE_MATCH_DATE)/' \
+	     -e 's/{{ .YOUR_RIMBLE_MATCH_ID }}/$(RIMBLE_MATCH_ID)/' \
+	     -e 's/{{ .YOUR_RIMBLE_API_KEY }}/$(RIMBLE_API_KEY)/' \
+	     ./match-winner.json.template > ./tmp/match-winner.json
```
tahardi, line 46: Previously, I'd tried updating the example code to fetch a matchid. Per your feedback, however, we expect the user to pass in the match id and date. This allows us to keep the actual example code (i.e., main.go) largely unchanged.

esports_data_from_rimble/Makefile
```diff
@@ -40,29 +40,36 @@ build: tmp/x.wasm
...
-	@sed -e 's/{{ .YOUR_RIMBLE_API_KEY }}/$(RIMBLE_API_KEY)/' ./team-kill-diff.json.template > ./tmp/team-kill-diff.json
+	@sed -e 's/{{ .YOUR_RIMBLE_MATCH_DATE }}/$(RIMBLE_MATCH_DATE)/' \
+	     -e 's/{{ .YOUR_RIMBLE_MATCH_ID }}/$(RIMBLE_MATCH_ID)/' \
+	     -e 's/{{ .YOUR_RIMBLE_API_KEY }}/$(RIMBLE_API_KEY)/' \
+	     ./team-kill-diff.json.template > ./tmp/team-kill-diff.json
 	@cat ./tmp/team-kill-diff.json | bky-as attest-fn-call >tmp/out.json
 	@echo "Output:"
 	@jq -r '.transitive_attested_function_call.claims.output | @base64d | fromjson' tmp/out.json
 
+.PHONY: fetch-match-id
```
tahardi, line 62: Added a convenience target for fetching a recent match id and date

esports_data_from_rimble/main.go
```diff
@@ -40,7 +40,7 @@ func getMatchDataFromRimble(
 	match, err := rimble.MakeMatchDataFromMatchesJSON(resp.Body)
 	if err != nil {
 		err = fmt.Errorf(
-			"making match given data '%s' and match ID '%s': %w",
+			"making match given date '%s' and match ID '%s': %w",
 			date,
```
tahardi, line 44: typo fix

esports_data_from_rimble/main.go
```diff
@@ -119,23 +119,16 @@ func matchWinnerFromRimble(inputPtr uint64, secretPtr uint64) uint64 {
...
 	MatchID  string
 	Date     string
-	MapName  string
 	Team1    string
 	Team2    string
 	KillDiff int
 }
 
-func TeamKillDifferenceOnMap(
+func TeamKillDifference(
```
tahardi, line 127: The one change I made to the example is that the "team kill diff" is now calculated for all games in a match as opposed to all games on a particular map. I chose to do this to limit the number of env vars the user has to setup. They now have to pass in api key, match id, and date. While they could also pass in a map name I figured this still version of kill diff still gets the point across and doesn't require another env variable

test/scripts/esports_data_from_rimble.txtar
```diff
@@ -5,7 +5,12 @@ cp stdout result-match-winner.json
 
 # [check] assert stdout contains correct match information
 exec jq -r '.transitive_attested_function_call.claims.output | @base64d | fromjson' result-match-winner.json
-cmp stdout expected-output-match-winner.json
+cp stdout output.json
+exec jq -e '.Success == true' output.json
+exec jq -e '.Error == ""' output.json
+exec jq -e '.Value.MatchID' output.json
+exec jq -e '.Value.Date' output.json
+exec jq -e '.Value.Winner' output.json
```
tahardi, line 13: Because the match data is subject to change I can only check for the expected field names not their actual values. I tried doing some fancy regex stuff I could not get commands with multi-line strings to work with testscript/txtar

.github/workflows/on-prs.yml
```diff
@@ -47,12 +47,32 @@ jobs:
         with:
           nix_path: nixpkgs=channel:nixos-24.11
 
+      - name: Fetch Recent Match ID
```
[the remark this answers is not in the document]
mwittie: Nice!

PR #67, opened 2025-07-11
[BKY-7043] Address CVEs

tahardi:
## Describe your changes
Unfortunately, the vulnerability deps are indirect deps introduced by the `hardhat` and `hardhat-toolbox` dependencies. Neither of which have updated their dependencies.

My understanding is you are generally not supposed to edit the `package-lock.json` file manually. I could use `override` in `package.json` to force all dependencies to use a newer version, but

1. This could lead to breaking changes
2. `package-lock.json` still lists the old versions which will cause Dependabot to still think we have vulnerable packages

So I chose to edit the dependency versions manually in `package-lock.json`. I was able to address all but one of the CVEs (it would appear that it needs the old version as `npm install` always reverts it). Also, I went ahead and updated `cookie` and `brace-expansion` to address the low CVEs as well since I was in there.

## Related Jira cards
[address cves](https://blocky.atlassian.net/browse/BKY-7043?atlOrigin=eyJpIjoiOTY3ZWE3OGI5OTEyNGJhOGE5ZTU4NWZkMmU5NTQ5MjUiLCJwIjoiaiJ9)

- [x] This PR is small.
    - Otherwise, justify here:

### DIFF
--- a/error_handling_on_chain/package-lock.json
+++ b/error_handling_on_chain/package-lock.json
@@ -1466,7 +1466,7 @@
         "@sentry/tracing": "5.30.0",
         "@sentry/types": "5.30.0",
         "@sentry/utils": "5.30.0",
-        "cookie": "^0.4.1",
+        "cookie": "^0.7.0",
         "https-proxy-agent": "^5.0.0",
         "lru_map": "^0.3.3",
         "tslib": "^1.9.3"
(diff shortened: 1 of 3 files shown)


tahardi:
Don't like this "manually edit package-lock.json" approach. Going to re-evaluate our use of hardhat and consider rewriting on-chain examples

PR #70, opened 2025-07-17
[BKY-6889] Test twap attest-fn-call

matteobjornsson:
## Describe your changes
This PR adds the `time_weighted_average_price/attest_fn_call` example to the test suite. 

## Related Jira cards
https://blocky.atlassian.net/browse/BKY-6889

## Notes for reviewers
we seem to be stretching the way we're using testscript here. Open to suggestions/discussion

- [ ] I have updated README files, if applicable.
- [ ] I have run `make pre-pr` and all checks have passed.
- [ ] I am merging into the intended base branch.
- [ ] This PR is small.
    - Otherwise, justify here:


time_weighted_average_price/attest_fn_call/Makefile
```diff
@@ -37,34 +37,36 @@ build: tmp/x.wasm
 
 .PHONY: prev
 prev: check
-	$(eval prev_ea := $(shell jq '.enclave_attested_application_public_key.enclave_attestation' tmp/prev.json | sed 's/\//\\\//g' ))
-	$(eval prev_ta := $(shell jq '.transitive_attested_function_call.transitive_attestation' tmp/prev.json | sed 's/\//\\\//g' ))
+	$(eval prev_ea := $(shell jq -r '.enclave_attested_application_public_key.enclave_attestation' tmp/prev.json))
+	$(eval prev_ta := $(shell jq -r '.transitive_attested_function_call.transitive_attestation' tmp/prev.json ))
```
matteobjornsson, line 41: some small updates to the make recipes since I am replicating them in the testscript and these seemed like they were doing some substitutions instead of using a different sed delimiter, which I think reduces the number of steps that can error and helps the reader follow more easily.

time_weighted_average_price/attest_fn_call/iteration-call.json.template
```diff
@@ -4,13 +4,13 @@
     "input": {
         "coin_id": "polygon-pos-bridged-weth-polygon-pos",
         "num_samples": 3,
-        "eAttest": VAR_EATTEST,
-        VAR_TATTEST
+        "eAttest": PREV_ENCLAVE_ATTESTATION,
+        "tAttest": PREV_TRANSITIVE_ATTESTATION,
```
matteobjornsson, line 8: I'm not sure why the eAttest field was always included but the tAttest field was conditional, where the end result was the same. I updated this file so they are treated the same.

matteobjornsson: I also updated the naming since I was already changing this part everywhere to something slightly more self-describing

time_weighted_average_price/attest_fn_call/Makefile
matteobjornsson: my understanding is this was escaping / characters because / is allowed in base64 encoding. I changed the sed delimiter below to | instead so all base64 characters are safe and we don't need to do any defensive escaping.

test/scripts/time_weighted_average_price_attest_fn_call.txtar
```diff
@@ -0,0 +1,70 @@
+# prepare iteration-call template for iteration 1 (bootstrap without previous attestations)
+exec sed -e 's|PREV_ENCLAVE_ATTESTATION|null|' -e 's|PREV_TRANSITIVE_ATTESTATION|null|' ./iteration-call.json.template
+cp stdout ./tmp/iteration-call.json
+
+# perform initial attest-fn-call
+stdin ./tmp/iteration-call.json
+exec bky-as attest-fn-call
+cp stdout tmp/prev.json
+exec jq -r '.transitive_attested_function_call.claims.output | @base64d | fromjson' tmp/prev.json
```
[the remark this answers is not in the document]
tahardi: I forgot to add, but consider asserting the success and error for each of this initial iterations (similar to how you save and assert output with `final-output.json`)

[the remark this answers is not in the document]
matteobjornsson: Good catch, likely a leftover from debugging (also the makefile flow does this to show the user the output). 

Honestly though, it is _quite_ helpful when debugging this test. For example, when an error occurs in one iteration, these statements print out 
```
        {
          "Success": false,
          "Error": "extracting priceSamples: previous run was an error: %!w(<nil>)",
          "Value": null
        }
```
But the test continues. I think your suggestions will address this "test carries on in a broken state" problem and we can remove the output printing. I'll take a look.

test/scripts/time_weighted_average_price_attest_fn_call.txtar
```diff
@@ -0,0 +1,70 @@
...
+
+# perform twap-call for final calculation
+stdin ./tmp/twap-call.json
+exec bky-as attest-fn-call
+cp stdout tmp/twap.json
+
+# assert the final output is successful
+exec jq -r '.transitive_attested_function_call.claims.output | @base64d | fromjson' tmp/twap.json
+cp stdout final-output.json
+exec jq -e '.Success == true' final-output.json
```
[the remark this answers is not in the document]
matteobjornsson: Updated

matteobjornsson:
@dlm I'm gonna merge this one in. Feel free to reach out if you have any issues here or want to discuss.

matteobjornsson:
confirmed that these changes do not break the twap on-chain example

PR #73, opened 2025-07-22
[BKY-6885] Combination attest-fn-call -> on-chain tests

tahardi:
## Describe your changes
The on-chain tests use hardcoded transitive attestations (TAs) for testing. Thus, updates to the TA format will not be caught by these tests. This PR adds "combination" tests that use the output transitive attestations from attest-fn-call tests as inputs to their on-chain counterparts. 

## Related Jira cards
[combo hello-world attest-fn-call -> on-chain example test](https://blocky.atlassian.net/browse/BKY-6885)

- [ ] I have updated README files, if applicable.
- [x] This PR is small.
    - Otherwise, justify here:

### DIFF
--- a/test/testscript.go
+++ b/test/testscript.go
@@ -194,6 +194,8 @@ func (e *TestscriptTest) Run(scriptFile string) {
 	testscript.Run(e.t, e.params)
 }
 
+// setEnvValueFromFile reads the contents of a file (filePath) and sets an
+// environment variable (envKey) to the rendered contents of the file
 func setEnvValueFromFileCmd(
 	cmdName string,
 ) func(*testscript.TestScript, bool, []string) {


test/examples_test.go
```diff
@@ -14,7 +14,7 @@ func TestCoinPricesFromCoingecko(t *testing.T) {
 	requiredEnvVars := []string{
 		"YOUR_COINGECKO_API_KEY",
 	}
-	NewTestscriptTest(t, projectDir).
+	NewTestscriptTest(t, projectDir, "").
```
tahardi, line 17: Now `TestscriptTest` expects a "save directory" where the `saveFile` command will save output files. Even if passed as a relative path it will get converted to an absolute path so the testscript is able to save the file to a location outside of its temporary work directory.

If no save directory is supplied, then the `saveFile` command simply prints a warning that the file was not saved and the test moves on without error. This behavior is so that we can reuse testscript scripts without requiring that they always write output files (we want to save the output for combination tests but not for normal tests).

test/examples_test.go
```diff
@@ -43,14 +43,41 @@ func TestErrorHandlingOnChain(t *testing.T) {
...
 
+func TestErrorHandlingCombined(t *testing.T) {
+	errorHandlingName := "error_handling_attest_fn_call"
+	errorHandlingDir := filepath.Join(examplesDir, errorHandlingName)
+	errorHandlingOnChainName := "error_handling_on_chain"
+	errorHandlingOnChainDir := filepath.Join(examplesDir, errorHandlingOnChainName)
+	saveDir := errorHandlingOnChainDir + "/tmp"
+
+	t.Run(errorHandlingName, func(t *testing.T) {
+		NewTestscriptTest(t, errorHandlingDir, saveDir).
```
tahardi, line 54: Here we run the `error_handling_attest_fn_call` examples test but give it a save directory of `error_handling_on_chain/tmp/`. So, when it uses the `saveFile` command it will write files to the error handling on-chain example directory

test/hardhat.go
```diff
@@ -36,7 +38,17 @@ func (h *HardhatTest) NPMInstall() *HardhatTest {
...
 }
 
+func (h *HardhatTest) SetEnv(key string, value string) *HardhatTest {
+	h.prevEnv[key] = os.Getenv(key)
+	err := os.Setenv(key, value)
+	if err != nil {
+		h.t.Fatalf("failed to set environment variable %s: %v", key, err)
+	}
+	return h
+}
```
tahardi, line 48: Added a function for setting environment variables before running the hardhat test. This is used to specify alternative TA files when running tests. If there was an existing env variable value, it is saved and restored later after the test is complete.

test/scripts/attest_fn_call.txtar
```diff
@@ -6,3 +6,4 @@ cp stdout out.json
 # [check] assert stderr and stdout expected values
 exec jq -r '.transitive_attested_function_call.claims.output | @base64d' out.json
 stdout 'Hello, World!'
+saveFile out.json attest-fn-call-out.json
```
tahardi, line 9: Here is the `saveFile` command in action. Recall, if no `saveDirectory` was specified then this will simply print a warning and continue on. This allows us to use this one script for both `TestAttestFnCall` and `TestAttestFnCallCombined`

test/examples_test.go
```diff
@@ -101,10 +128,34 @@ func TestHelloWorldOnChain(t *testing.T) {
 		Run("--grep", "Local")
 }
 
+func TestAttestFnCallCombined(t *testing.T) {
+	attestFnCallName := "attest_fn_call"
+	attestFnCallDir := filepath.Join(examplesDir, attestFnCallName)
+	helloWorldOnChainName := "hello_world_on_chain"
```
tahardi, line 134: Should I update `hello_world_on_chain` to be `attest_fn_call_on_chain`?

time_weighted_average_price/on_chain/test/user.ts
```diff
@@ -52,9 +52,15 @@ describe("Local Tests", function () {
 
     it("Verify attested TWAP in User contract", async () => {
         // given
-        const evmLinkData: EVMLinkData = loadEVMLinkData("../inputs/twap.json");
+        const taFile = process.env.TA_FILE || "../inputs/twap.json";
+        const evmLinkData: EVMLinkData = loadEVMLinkData(taFile);
         const {userContract} = await loadFixture(deployUser) as UserContract;
 
+        function anyFloat(x: unknown): boolean {
```
tahardi, line 59: @matteobjornsson copied these changes from your PR#71

test/examples_test.go
```diff
@@ -171,3 +222,31 @@ func TestTWAPOnChain(t *testing.T) {
 		NPMInstall().
 		Run("--grep", "Local")
 }
+
+func TestTWAPCombined(t *testing.T) {
```
tahardi, line 226: is the "combined" name fine or is there a better way to describe these combination tests?

test/hardhat.go
matteobjornsson: @tahardi you should be able to use the test framework to set the environment rather than mutating the os environment with [t.SetEnv](https://pkg.go.dev/testing#T.Setenv)

test/examples_test.go
```diff
@@ -101,10 +128,34 @@ func TestHelloWorldOnChain(t *testing.T) {
...
+			CopyFile("fn-call.json").
+			Run(filepath.Join(scriptDir, attestFnCallName+".txtar"))
+	})
+
+	t.Run(helloWorldOnChainName, func(t *testing.T) {
+		NewHardhatTest(t, helloWorldOnChainDir).
+			NPMInstall().
+			SetEnv("TA_FILE", "../tmp/attest-fn-call-out.json").
+			Run("--grep", "Local")
+	})
```
matteobjornsson, line 152: As noted in another comment, you should be able to use t.SetEnv here, and the test framework will handle isolating/resetting the env values
```go
	t.Run(helloWorldOnChainName, func(t *testing.T) {
		// set the env directly in the test framework that's wrapping the hardhat test
		t.SetEnv("TA_FILE", "../tmp/attest-fn-call-out.json")
		
		NewHardhatTest(t, helloWorldOnChainDir).
			NPMInstall().
			Run("--grep", "Local")
	})
```

test/examples_test.go
matteobjornsson: seems fine to me. Could use "compatibility" like in zpr or you could name it something like TestTWAPExampleDependencies or TestTWAPIOCoupledExamples

matteobjornsson: dependencies is kind of an overloaded term though

test/scripts/attest_fn_call.txtar
matteobjornsson: normally I'd say we should make a new testscript for a new test, but this seems like a totally fine concession given how complex the testscript is.

test/testscript.go
```diff
@@ -17,9 +17,14 @@ type TestscriptTest struct {
...
 	setupFuncs []func(*testscript.Env) error
 	projectDir string
+	saveDir    string
 }
 
-func NewTestscriptTest(t *testing.T, projectDir string) *TestscriptTest {
+func NewTestscriptTest(
+	t *testing.T,
+	projectDir string,
+	saveDir string,
```
matteobjornsson, line 26: this seems to strongly impose a concept on the testscript test famework that is not used by most examples. I'd be interested in seeing if we can either achieve this on a case-by-case basis at a lower level, or go the other way and come up with another higher level test setup that defines input/output relationship dependencies, eg `NewIOCoupledTest` that gives structure to the expectations about what sort of dependencies are expected to be provided at each step by the previous step

matteobjornsson: We could then feed it an entire test function per step or something like that

test/testscript.go
```diff
@@ -153,41 +163,72 @@ func (e *TestscriptTest) Run(scriptFile string) {
...
-	if neg {
-		ts.Fatalf("unsupported: ! setEnvValueFromFile")
-	}
-	var srcRelPath string
-	if len(args) != 2 {
-		ts.Fatalf("usage: setEnvValueFromFile envKey filePath")
-	}
-	srcRelPath = args[1]
-	envKey := args[0]
+func saveFileCmd(
```
matteobjornsson, line 172: I can see the need for some supporting command written in go since testscript is so limited. I'm not sure about the baked in "saveDir" value. I sketched out another idea here https://github.com/blocky/attestation-service-examples/compare/feature/combo-test-bky-6885...temp/discuss-pr, maybe we should meet up and discuss?

test/testscript.go
```diff
@@ -153,41 +163,72 @@ func (e *TestscriptTest) Run(scriptFile string) {
...
-		"setEnvValueFromFile": cmdSetEnvValueFromFile,
+		"saveFile":            saveFileCmd("saveFile", e.saveDir),
+		"setEnvValueFromFile": setEnvValueFromFileCmd("setEnvValueFromFile"),
 	}
 	testscript.Run(e.t, e.params)
 }
 
-// setEnvValueFromFile envKey filePath
-// setEnvValueFromFile reads the contents of a file (filePath) and sets an
-// environment variable (envKey) to the rendered contents of the file
```
matteobjornsson, line 163: not a van of the usage comments?

matteobjornsson (changes requested):
I think this PR is great, but I think we are a bit between abstraction layers and would like to move it one way or another (either handle combo-test specific things at a lower level or put together another test type specific to handling inter-example dependencies). I left some comments that articulate the idea but I figure it would be easiset to meet up to discuss. LMK!

test/examples_test.go
```diff
@@ -101,10 +128,34 @@ func TestHelloWorldOnChain(t *testing.T) {
...
 }
 
+func TestAttestFnCallCombined(t *testing.T) {
+	attestFnCallName := "attest_fn_call"
+	attestFnCallDir := filepath.Join(examplesDir, attestFnCallName)
+	helloWorldOnChainName := "hello_world_on_chain"
+	helloWorldOnChainDir := filepath.Join(examplesDir, helloWorldOnChainName)
+	saveDir := helloWorldOnChainDir + "/tmp"
+
+	t.Run(attestFnCallName, func(t *testing.T) {
```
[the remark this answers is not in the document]
matteobjornsson: Update (reiterating what I messaged you in slack): 

Based on my understanding, the go testing framework is pretty baked into the testscript library. It waits until the parent test object finishes before executing any deferred/cleanup functions (testscript finishes, parent T finishes, testscript deferred functions, parent T deferred functions), so we can't use the defer/cleanup mechanisms to continue execution at the same test/subtest level.

It seems that if we want control over the testscript lifecycle I think we'd need to hand it something that fulfills the testscript.T interface
```go
// T holds all the methods of the *testing.T type that
// are used by testscript.
type T interface {
	Skip(...any)
	Fatal(...any)
	Parallel()
	Log(...any)
	FailNow()
	Run(string, func(T))
	// Verbose is usually implemented by the testing package
	// directly rather than on the *testing.T type.
	Verbose() bool
}
```
in a way that manages its relationship to the parent `*testing.T` object. 

For now, using `t.Run` seems to be on ok compromise.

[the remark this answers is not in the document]
tahardi: Awesome, thanks for figuring this out!

test/examples_test.go
tahardi: Thanks, removed environment setup from `HardhatTest` and use `testing.T` now instead

test/examples_test.go
tahardi: Im just going to leave combined for now. Works well enough and I don't want to get held up on naming

test/live/live_test.go
```diff
@@ -52,17 +53,64 @@ func TestLiveErrorHandlingAttestFnCall(t *testing.T) {
...
-		CopyFile("successFunc.json").
-		CopyFile("errorFunc.json").
-		CopyFile("panicFunc.json").
+		CopyProjectFile("successFunc.json").
+		CopyProjectFile("errorFunc.json").
+		CopyProjectFile("panicFunc.json").
 		Run(filepath.Join(scriptDir, projectName+".txtar"))
 }
 
+func TestLiveErrorHandlingCombined(t *testing.T) {
```
tahardi, line 67: Forgot to add live versions of the combo tests the first time around. Notice that I left out TWAP because it's templates are actually hardcoded to only work with the local server. There is a card for adding twap live tests so that will be handled in a separate PR

